PR #435, opened 2015-01-13
[fix] Updated country codes as per spec

alex-deaconu:
@stevesherrie please review. Thanks!


stevesherrie:
:+1: I guess I should line up a CH update for store also.

alex-deaconu:
@stevesherrie I've created v1.0.4 with this latest commit. Thanks.

PR #439, opened 2015-01-15
[fix] switchCompany called with a company id

alex-deaconu:
Instead of calling switchCompany with an object, it is called with an ID
It retrieves the object itself

@stevesherrie previous fix had timing issues where the route was changed (by adding cid) before the new company info was retrieved. This fix retrieves the new company info before switching the cid.

Please review. Thanks.


stevesherrie:
:+1:

PR #445, opened 2015-01-22
[feature] Added new company state service

alex-deaconu:
Service split out from userstate, deals with company part
To prevent breaking changes to apps, userstate still proxys function calls to company state

Simplified selectedCompanyUrlHandler service, by moving some logic to companystate
Split out objectHelper service for handling object copy, clear, follow

CH loading logic now waits for Company and Selected Company to load before triggering the user.authorized message


alex-deaconu:
@olegrise first iteration of this implementation, if you'd like to take a look and see if you agree with the changes.

Still working on unit tests for the new service and the selectedCompanyUrlHandler. Thanks.

dist/js/common-header.js
```diff
@@ -3928,6 +3901,117 @@ angular.module("risevision.common.geodata", [])
...
+    }
+    
+    var _init = function () {
+      var deferred = $q.defer();
+      
+      //populate userCompany
+      getCompany().then(function(company) {
+        objectHelper.clearAndCopy(company, _state.userCompany);
+                
+        return _switchCompany(pendingSelectedCompany);
```
olegrise: @alex-deaconu I would add a little optimization here to avoid calling getCompany API twice for the same company:
```
if (company.companyId !== pendingSelectedCompany) {
  return _switchCompany(pendingSelectedCompany);
} else {
  _companyState.resetCompany();
}
```

dist/js/common-header.js
```diff
@@ -3928,6 +3901,117 @@ angular.module("risevision.common.geodata", [])
...
+      })
+      .then(null, function() {
+        _companyState.resetCompany();
+      })
+      .finally(function () {
+        pendingSelectedCompany = null;
+        
+        deferred.resolve(null);
```
olegrise: @alex-deaconu I wonder how well this `deferred.resolve(null);` is going to play with this line https://github.com/Rise-Vision/common-header/pull/445/files#diff-2ec175c12c8011573203566f05849457R3938

It might be more correct in this situation to call `.then(_switchCompany(pendingSelectedCompany))` instead of using `return`.

olegrise:
@alex-deaconu nice work! Just couple of comments.

dist/js/common-header.js
alex-deaconu: @olegrise actually, `_switchCompany` returns a promise, so it does wait for that promise to complete before executing `.finally`. Thanks.

dist/js/common-header.js
alex-deaconu: @olegrise I agree this code is not very clear. 

In `_switchCompany` I check the `companyId` and call `resetCompany` if it is blank. Also, if the `getCompany` call fails, the promise will fail and the `resetCompany` will be called in `_init`.

However, my logic is the following: 
If the user is in their sub-company, and adds a cid that fails to produce a valid company, I didn't want to `resetCompany` on them in that case. The call will just fail silently (updated console log) and not change the state.

Respectively, if switchCompany is called with a blank companyId, it will just call `resetCompany`.

src/js/svc-companystate.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+    
+    var _init = function () {
+      var deferred = $q.defer();
+      
+      //populate userCompany
+      getCompany().then(function(company) {
+        objectHelper.clearAndCopy(company, _state.userCompany);
+        
+        return _switchCompany(pendingSelectedCompany);
```
olegrise: @alex-deaconu there are 5 use cases here:
1) pendingSelectedCompany is null. This is covered.
2) pendingSelectedCompany is invalid Company ID. This is covered as well.
3) pendingSelectedCompany is valid Company ID, but user does not have rights to access it. This is the same case invalid Company ID. Covered.
4) pendingSelectedCompany is valid Company ID and it's a sub-company. Covered.
5) pendingSelectedCompany is the same as user Company ID. This is covered, **but you are making 2 calls to the Store API to load the same company.** Not a big deal, however I believe this is the most common use case. Just wanted to clarify on my previous comment about optimization.

dist/js/common-header.js
olegrise: @alex-deaconu sounds good. Thanks for the clarification.

src/js/svc-companystate.js
alex-deaconu: @olegrise I see what you mean. I was missing the point because we don't call `switchCompany` with the userCompanyId, `resetCompany` is used for that. Also the querystring `cid` parameter is `null` when the `selectedCompanyId = userCompanyId`. So technically scenario 5 would never happen.

I do however agree with you this needs to be covered in the code to prevent the extra API call. Will adjust accordingly. Good catch, thanks!

alex-deaconu:
@olegrise I've added new tests for the code I modified. Also, as per your suggestion I added a check for the userCompanyId triggering an extra API call, with corresponding test coverage.

Please review, once this is ready I'll squash the commits. 

Thanks!

olegrise:
:+1:

alex-deaconu:
@olegrise fixed issue https://github.com/Rise-Vision/common-header/issues/448 by copying the $scope object and removing the fields from there. 

Also updated the unit tests to verify this.

Please review. Thanks.

olegrise:
:+1:

PR #447, opened 2015-01-23
Fix bug in adjustItemQuantity and refactor shoppingCart service

olegrise:
- fix bug in adjustItemQuantity.
- refactored shoppingCart service. Removed itemsMap. Use of itemsMap was inconsistent and was causing more problems than providing benefits.

@alex-deaconu please review


alex-deaconu:
Thanks for fixing. PR looks good, nice simplification :+1: .

PR #453, opened 2015-02-09
[feature] Company Address UX and added Timezone

alex-deaconu:
Note: the dialog that is being opened needs to be set to `lg` size. Not sure why the call had `companId, size` parameters since they weren't being used. Hardcoded `size` for now, I don't see the need for a parameter.


olegrise:
:+1:

PR #457, opened 2015-02-09
[fix] Adding integer parser for timezone

alex-deaconu:
Core API expects timezone to be integer
Dropdown value returns String
Created directive that will automatically convert value to int
Value will default to 0 if invalid
http://stackoverflow.com/questions/15072152/angularjs-input-model-changes-from-integer-to-string-when-changed

@olegrise this fixes an issue with saving new Companies; sorry for catching so late. Please review. Thanks.


olegrise:
:+1:

PR #458, opened 2015-02-12
[fix] Persisting Authorization promise

alex-deaconu:
Persisting Authorization promise
Clearing promise when user state is reset

Bower dependency resolution
CSS build results


alex-deaconu:
@olegrise @stevesherrie as per yesterday's discussions I'm making the Authorization promise persistent. What this means is that even if an App calls `userState.authenticate()` on startup, the User and Company API calls will only happen once. However that promise get reset on `_resetUserState` so the calls happen again.

Please take a look and see if you see any issues with this change. Thanks!

olegrise:
@alex-deaconu I like your idea in general. but [this](https://github.com/Rise-Vision/common-header/blob/fix/persist-auth-promise/src/js/svc-userstate.js#L176) could be a problem. This code refreshes token every 55 minutes.

olegrise:
@alex-deaconu also take a look at [this](https://github.com/Rise-Vision/common-header/blob/fix/persist-auth-promise/src/js/svc-userstate.js#L110). It might stop working as well.

alex-deaconu:
Both good points @olegrise. Do you think this could be ameliorated by separating the User/Company API calls from the token refresh calls? Basically from here: https://github.com/Rise-Vision/common-header/blob/fix/persist-auth-promise/src/js/svc-userstate.js#L281

This way both the calls you're pointing out will get a new token/refresh the token, but we would still get the benefit of promise caching if 
` if(!_state.user.username || !_state.profile.username ||
                     _state.user.username !== oauthUserInfo.email) {`
turns out to be true (which is where the duplicate calls come in).

Thanks.

alex-deaconu:
@olegrise actually disregard that. I need to think about this one some more. Will come up with an improved version and let you know. Thanks for the review!

olegrise:
@alex-deaconu sounds good :)

alex-deaconu:
@olegrise a few improvements on this PR. First off, I've moved the promise caching out of `_authorize` into `authenticate`. The idea is that `authenticate` is a public function, that's cached; `_authorize` is an internal function. However, forcing `authenticate` (`forceAuth === true`) resets the promise.

Can you take a look and let me know your thoughts? Thanks!

src/js/svc-userstate.js
```diff
@@ -117,17 +118,16 @@
...
+        if (token !== _state.userToken) {
           //token change indicates that user either signed in, or signed out, or changed account in other app
           $window.location.reload();
         } else if (_state.userToken) {
           //make sure user is not signed out of Google account outside of the CH enabled apps
-
-          // [AD] Do not show spinner for this
-          // $loading.startGlobal("risevision.user.authenticate"); //spinner will be stop inside authenticate()
-          authenticate(false).finally(function () {
+          _authorize(true).finally(function () {
```
olegrise: @alex-deaconu why force authentication `_authorize(true)`? Let's say user has Store open on 2 tabs. He signs out in the first tab, then clicks on the second tab. What is the expected outcome in this situation? I don't think Store should ask user to sign in again.

alex-deaconu: @olegrise good question. `authenticate` calls `_authorize(!forceAuth)` in this manner, so the value is reversed. It won't force the user to sign in again. Thanks.

alex-deaconu:
@olegrise this next commit fixes issues with the Registration scenario. 

Also, it removes some un-needed/unused code. I didn't find any references in Store for the removed `userStateReady` service either. Can you please confirm? 

Thanks!

olegrise:
I don't see any references to the `userStateReady` in the Store.

src/js/svc-userstate.js
olegrise: @alex-deaconu right!

src/js/controllers/ctr-registration-modal.js
```diff
@@ -75,7 +75,7 @@ angular.module("risevision.common.header")
 
           action.then(
             function () {
-              userState.authenticate(false).then()
+              userState.refreshProfile().then()
                 .finally(function () {
```
olegrise: @alex-deaconu have you tested this change? From the first glance it looks like you need to load user company information i.e. `companyState.init` after user registration and `userState.refreshProfile()` does not do that.

alex-deaconu: @olegrise yes, you're right. However, I moved `companyState.init` to `userState.refreshProfile()` to address this. Thanks.

dist/js/common-header.js
```diff
@@ -3903,41 +3892,41 @@ angular.module("risevision.common.geodata", [])
             userInfoCache.removeAll();
```
olegrise: in `_resetUserState` you do `_authenticateDeferred = null;`. Could this be a problem?

dist/js/common-header.js
```diff
@@ -3740,6 +3723,7 @@ angular.module("risevision.common.geodata", [])
       };
 
       var _resetUserState = function () {
+        _authenticateDeferred = null;
         objectHelper.clearObj(_state.user);
```
olegrise: @alex-deaconu let's say 
- application calls `authenticate.then(...)`. It receives pointer to the promise object A.
- then another part of application calls `_resetUserState` (possible?). It clears `_authenticateDeferred `. 
- then `authenticate.then(...)` is called again. `_authenticateDeferred ` is recreated and pointer to the promise object B is returned.
- then inside `authenticate` function it calls `_authenticateDeferred.resolve();`. Now promise B is resolve, but promise A is not.

dist/js/common-header.js
alex-deaconu: @olegrise good point. Will revise momentarily. Thanks.

dist/js/common-header.js
alex-deaconu: @olegrise interesting. `_resetUserState` is only called when the user logs out or when we want to do forced authentication. However, I do agree this could lead to a problem of un-resolved promises. Will revise to use a local copy of the promise. Thanks.

alex-deaconu:
@olegrise I've addressed the points you've brought up in the latest round of review. To summarize:

I've improved the way `authenticate` deals with the cached promise. The promise is reset by `_resetUserState`, if the User logs out, or registers, is always recreated if it is blank.

In addition, the `authenticate` now uses a local variable for the promise, such that if the promise does happen to get reset, calls which are waiting for the "old" promise resolution will be fulfilled. 

Let me know your thoughts on the above. Thanks!

dist/js/common-header.js
```diff
@@ -3894,50 +3876,61 @@ angular.module("risevision.common.geodata", [])
...
-              //  _clearUserToken();
-              authenticateDeferred.reject(msg);
-              objectHelper.clearObj(_state.user);
-              $loading.stopGlobal("risevision.user.authenticate");
-            }
-          });
+                  $log.debug("Authentication Error: " +
+                    authResult.error);
+                  _authenticateDeferred.reject(
+                    "Authentication Error: " + authResult.error);
```
olegrise: @alex-deaconu did you miss this `_authenticateDeferred `?

alex-deaconu:
@olegrise good catch, fixed. Thanks.

olegrise:
@alex-deaconu looks good so far.

olegrise:
@alex-deaconu just tested it with the Store. If I sign out, then I can't sign it back. I'm using Store from master.

alex-deaconu:
@olegrise oh.. thanks for bringing this up. I'll try to replicate and identify the issue. Thanks!

alex-deaconu:
@olegrise the issue you have brought up re: sign out/in has been resolved. Also fixed an issue with signing out the user when the tab is switched (latest commit).

Do you think this is ready for merge? Thanks.

olegrise:
Looks goods. I tested with the Store - no issues. Good job @alex-deaconu.

I'm looking forward to clean up the Store code and start using userState.authenticate() everywhere.

alex-deaconu:
@olegrise sounds good. Your review/feedback was invaluable for getting this PR through. Thanks!

PR #460, opened 2015-02-17
[fix] Redirect url & state logic

alex-deaconu:
Add trailing slash (/) to redirect URL and remove from path
- issue caused by the fact that return urls are registered with the trailing /

Fix state logic
- append path/state to url all the time


alex-deaconu:
@olegrise A couple of fixes on the PR from last week. Please review. Thanks!

olegrise:
:+1:

PR #464, opened 2015-02-18
[fix] Updating redirect logic

alex-deaconu:
Redirect now checks if the path/search are applied
Append current hash to redirect URL

@olegrise some further optimizations for the CH redirect logic. This PR should resolve some issues I've been seeing with deployed apps where the hash containing the user token is lost.

Please review. Thanks!


olegrise:
:+1:

Issue #470, opened 2015-02-26
Time Zone is saving, but not showing in CH


robb-price:
<b>How it works now:</b>
Change the Time Zone in your Company Settings, notice that regardless of what you select, the dropdown stays as "Select Time Zone". Save and reopen the Company Settings, notice that it still says "Select Time Zone".

Go into the RVA, and the selection you saved is what is set for your Company.

For further clarification: https://dl.dropboxusercontent.com/u/8237394/Screencasts/2015-02-26_1009.swf

alex-deaconu:
@robb-price I can't replicate this issue in the Displays App. Timezone is showing and editing correctly.

@stevesherrie does Store define its own TIMEZONES constant anywhere? 

Thanks.

brianloosbrock:
@stevesherrie @alex-deaconu We can park this for now and come back to it. Thanks

alex-deaconu:
@robb-price @brianloosbrock this issue should be fixed with the new CH version in the Store. Thanks!

robb-price:
@alex-deaconu @brianloosbrock Looks like it is corrected.

Verified and Closed.

PR #473, opened 2015-02-26
[fix] Using ng-repeat for timezone options

alex-deaconu:
Seems like ng-options doesn't work well with angular 1.2.*

Fix for issue #470.


olegrise:
:+1:

PR #477, opened 2015-03-11
[fix] Set or clear hash parameter

alex-deaconu:
After the User returns to the app from the Authorization page, the hash parameter contains the token. This needs to be cleared regardless of whether or not there's a hash on the page.


olegrise:
:+1: Makes sense.

PR #479, opened 2015-03-11
[fix] Using sub-domain for cookie storage

alex-deaconu:
only required for appspot.com urls

Somehow on appengine apps the cookie doesn't persist if we set the domain to `appspot.com`. It requires a sub-domain to be set, ie. `rva-test.appspot.com`.

@olegrise any thoughts/suggestions on the above? Otherwise, please review. Thanks!


olegrise:
@alex-deaconu sounds strange, but I don't have a better idea how to resolve this. Could you add a comment to the code so we know why it's there?

alex-deaconu:
@olegrise I'm assuming google doesn't want cookies to be shared across appspot.com sub-domains. I'd say it's a good security measure. Added comment to describe the situation, as requested. Thanks!

PR #480, opened 2015-03-12
Fix/remove hash from shopping cart link

rodrigopavezi:
@alex-deaconu please review. cheers


alex-deaconu:
@rodrigopavezi would removing the hash cause any issues with other apps that use the Shopping Cart? Thanks.

rodrigopavezi:
@alex-deaconu yes, if they don't implement html5 push state. Witch other app is using the shopping cart? What they could do is to set the .value("STORE_URL", ""); to .value("STORE_URL", "#");.

alex-deaconu:
@rodrigopavezi okay. @olegrise any thoughts on the above breaking change? Also, do you know where else the Shopping Cart is used outside the store? Thanks.

rodrigopavezi:
@alex-deaconu sorry, I am confusing things. I think it will work because the shopping-cart is a store page, right? Store is implementing push state. So if other app uses the cart it would have a value("STORE_URL", "http://store.risevision.com/");. Then the link would be http://store.risevision.com/shopping-cart which is right.

olegrise:
It doesn't look like this particular change will affect other apps because STORE_URL specifically points to the Store app. However I don't feel too comfortable which these changes in general because I'm new to HTML5 push state.

There are other places where `#` is used:

https://github.com/Rise-Vision/common-header/blob/master/src/js/dtv-common-header.js#L53
https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-userstate.js#L65
https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-userstate.js#L340
https://github.com/Rise-Vision/common-header/blob/master/src/templates/app-nav-buttons-menu.html (3 places)
https://github.com/Rise-Vision/common-header/blob/master/src/templates/signup-modal.html#L15

@rodrigopavezi is it safe to leave them as is?

rodrigopavezi:
@olegrise good point.

https://github.com/Rise-Vision/common-header/blob/master/src/js/dtv-common-header.js#L53
This would bother as it is the default  CH links so apps would change it.

https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-userstate.js#L65
https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-userstate.js#L340
This two, I am not sure. I don't know where they are used. @alex-deaconu can you help us with that?

https://github.com/Rise-Vision/common-header/blob/master/src/templates/app-nav-buttons-menu.html
This links will change so I believe the # won't be there.

https://github.com/Rise-Vision/common-header/blob/master/src/templates/signup-modal.html#L15
This does not have to change. It should continue to work.


I am more concerned about the ones from svc-userstate. 

cheers

alex-deaconu:
@rodrigopavezi if the hash exists it is stored by the CH in a `state` variable which is returned one the user redirects back to the app from the authentication page. This `state` should work fine even without a hash parameter, however, it will append a /# at the end of any URL (I can fix that if needed). Thanks.

rodrigopavezi:
@alex-deaconu do you think it could break an app that uses #? cheers

alex-deaconu:
@rodrigopavezi what are you referring to?

rodrigopavezi:
@alex-deaconu if an app that uses comman-header and does not use html5 push state. Could this app angular not understand the request because it does not contain a #?

alex-deaconu:
Yes, if the app doesn't support html5 push state and you're adding a path directly to the URL without using the #, it will break that app (ie. http://displays.risevision.com/display/WNX325MXWMJY).

rodrigopavezi:
@alex-deaconu agreed. then applying the fix on the CH state which will remove /# would break those apps right?

alex-deaconu:
@rodrigopavezi I think you misunderstood what I was proposing. The fix would remove the hash entirely if it is not needed/present in the `state`. This should function correctly. To confirm, we're talking about http://displays.risevision.com vs http://displays.risevision.com/#/ which are both valid URLs. If the `state` had a hash parameter, it would obviously be appended to the end of the URL.

rodrigopavezi:
@alex-deaconu thanks. so we are fine. cool.

rodrigopavezi:
@alex-deaconu is it okay with you if you make the change you stated above? I can try doing that if it is going to take too much of your time. I will just need some help from you. cheers

alex-deaconu:
@rodrigopavezi the last commit you added is incorrect. `state.p` already contains the path. `state.s` contains the querystring. What you're doing is adding the `hash` parameter after the querystring without a `#`. That doesn't make sense. Please remove it. 

The fix I was talking about is in this line (2 lines down): 
`          $location.path(state.u ? state.u : "");`

If `state.u` is blank, instead of setting the angular `$location.path` set `$window.location.hash = ""` (blank).

As this fix is not urgent, I can make the necessary change (and test it) at a later date. Will add an issue for it. Thanks.

rodrigopavezi:
@alex-deaconu Thanks for the heads up. I have reverted it back. I will leave this with you. Could you please review the rest and let me know if can merge it?

cheers

alex-deaconu:
@rodrigopavezi can you re-build the dist/common-header.js file? Otherwise looks good. Don't forget to squash before merge. Thanks!

rodrigopavezi:
@alex-deaconu sure, sorry about that. cheers

alex-deaconu:
Added #481 for removing the hash. Thanks.

alex-deaconu:
@rodrigopavezi LG :+1: . Thanks for the squash!

olegrise:
@rodrigopavezi @alex-deaconu is it safe to use the latest version of CH with the current version of Store? The reason I ask is because I need to make few changes to the CH and use it in the Store right away. If we can't use the latest version of CH until Store implements html5 push state, then should we roll this PR back?

rodrigopavezi:
@olegrise I believe you can use this version but you will have to change the STORE_URL on the store config files. It will have to be STORE_URL = "#/".  

ref: https://github.com/Rise-Vision/common-header/blob/master/src/js/controllers/ctr-shoppingcart.js#L18

cheers

olegrise:
Thanks @rodrigopavezi. You are right. It's less work that rolling this PR back.

PR #488, opened 2015-03-20
Changed check for redirectToRoot option.

rodrigopavezi:
@alex-deaconu could you please review? What do you think? This fixed the issue with store so we don't need to move CH to the index.html cheers


alex-deaconu:
@rodrigopavezi looks good. @olegrise any objections? Thanks.

olegrise:
Looks good.

rodrigopavezi:
@alex-deaconu is there okay with you if I merge it now and create a new version? cheers

alex-deaconu:
@rodrigopavezi something's weird with your PR, more specifically the second commit. Can you squash and rebase so we can only see the first commit's changes? Thanks.

PR #489, opened 2015-03-20
Add surrencyService with tests.

olegrise:
@alex-deaconu please review. 
[Trello card](https://trello.com/c/jVaYBmyl/383-store-currency-changes-server-side)
The idea is that currencies are loaded from the Store API vs. being hard-coded. The currencyService is mostly going to be used by the Store, but I realized that Storage might need it as well because they are going to place "Add To Cart" button right on their home page. That is the reason for currencyService being added to the CH, not the to the Store.


alex-deaconu:
@olegrise reviewed. Added a couple of comments/suggestions in the code.

I'm still working on the Store PR, will finish off on Monday as discussed. 

Thanks.

olegrise:
@alex-deaconu I've made few more changes. Please review. This has been tasted by QA on stage env.

alex-deaconu:
@olegrise looks good :+1:. Thanks.

PR #498, opened 2015-03-31
[feature] Update Countries list to use service

alex-deaconu:
@olegrise as per discussion here: https://github.com/Rise-Vision/common-app-components/pull/4, making the countries.list load a prerequisite for showing the controllers.

Please review. Thanks!


olegrise:
:+1:

Issue #487, opened 2015-03-20
When in a sub-company, clicking an app in the App Launcher places you in the account where you are a registered user


mfisher11:
Steps to reproduce:
1. Go into a sub-company
2. Open the App Launcher
3. Click on an App (confirmed with editor, store, storage, displays, and schedules)

Result:
Content (appropriate for the App) is listed for the account where you are a registered user

How it should work:
Content (appropriate for the App) is listed for the current sub-company.

Link to Presentation:
N/A

OS:
Xubuntu 14.04 and OSX

Chrome Ver:
Version 41.0.2272.89 (64-bit)

alex-deaconu:
@alanclayton @mfisher11 do we expect all links to behave in this manner. This includes `Editor`, `Store`, `Storage`, `Displays`, `Schedules`, `Support`.

However, should `Developer`, `Documentation`, `About Us` behave in a similar manner (load the CH with the sub-company selected)? I don't believe the selected company is relevant in those apps.

As for `Community` it doesn't use the CH so I'll leave it as is.

Thanks.

mfisher11:
@alex-deaconu @alanclayton  This kind of decision is above my pay grade, but I would think that it should apply to everything that is relevant (the first group) and not necessarily the second.

It would also depend on the amount of work:  if you have to change every item, then just do the ones that matter; but if a single change affects all of them, I'd go ahead and do it for all without carving out exceptions unless it breaks something in dev, doc, about...

alanclayton:
@alex-deaconu Excluding `Community` I think we should be consistent and load CH with the sub-company selected for each App. 

Is it much more work to do that vs. only doing it for specific Apps?

alex-deaconu:
@alanclayton @mfisher11 not really more work. I was just thinking of the user may see the sub-company notification when accessing those other apps and wonder what's up.

I'll add it to all for now, and we can easily remove it if needed.

Thanks!

robb-price:
Confirmed at http://1-08-005.rvauser2.appspot.com/#/ that this is working for everything except Support: https://dl.dropboxusercontent.com/u/8237394/Screencasts/2015-04-06_1429.swf

robb-price:
Verified that Support is now functioning correctly at http://1-08-005.rvauser2.appspot.com/.

robb-price:
Verified and Closed.

PR #500, opened 2015-04-01
[fix] Add cid to nav menu links

alex-deaconu:
Update cid when the company changes

Fix for issue #487.


dist/js/common-header.js
```diff
@@ -5089,6 +5089,34 @@ function (loadFastpass, userState) {
   };
 }]);
 
+angular.module("risevision.common.header")
+.directive("linkCid", ["userState",
+function (userState) {
+  return {
+    link: function($scope, ele, attr){
+      var updateLinkCompanyId = function(companyId) {
+        var index = attr.href.indexOf("?cid=");
```
olegrise: @alex-deaconu what if `cid` is not the first parameter? Could be `"?cid="` or `"&cid="`. Possible?

alex-deaconu:
@olegrise updated the parsing as per your query. It should handle `?` and `&` correctly. Thanks!

olegrise:
:+1:  @alex-deaconu I just realized that my comment was irrelevant to the way you use the directive. Sorry.

alex-deaconu:
@olegrise no worries. The extra check doesn't hurt.

Issue #502, opened 2015-04-06
Clicking on the "Already a Rise Vision User" Sign in link in the modal does not log you in correctly.


robb-price:
<b>How it works now:</b>
1. Go to http://1-08-005.rvauser2.appspot.com/#/, click on the Sign Up Free button, click the Sign In link next to "Already a Rise Vision User" text. 
2. Notice that it starts to log in, then asks you to select the account to log in as again. When you click the account, there is nothing in the CH except the navigation buttons.

For further clarification: https://dl.dropboxusercontent.com/u/8237394/Screencasts/2015-04-06_1440.swf

alex-deaconu:
@robb-price this happens with the current RVA version (rva.risevision.com) so it's an existing issue. Will fix, however I'd like to do so independently of this release. @alanclayton FYI.

Thanks.

robb-price:
Verified at http://1-08-005.rvauser2.appspot.com/

robb-price:
Verified and Closed.

PR #504, opened 2015-04-06
[fix] Updated login button logic

alex-deaconu:
Calling userState.authenticate rather than triggering via event.

Fix for #502.


olegrise:
:+1:

PR #506, opened 2015-04-07
[chore] Add jsbeautifier task and results

alex-deaconu:
No code changes... just formatting results.


olegrise:
:+1: I like this change. My only concern is that CircleCI build will re-format the source code which it useless.

alex-deaconu:
@olegrise yes, this task has become pretty common in my projects. I suggest it for the Store as well. 

Agreed re: running it in CircleCI is a waste, but if I took it out of the build, it would probably never get run on its own.

Thanks.

olegrise:
@alex-deaconu agree with both points.

PR #509, opened 2015-04-15
[chore] Split out rvToken and baseDomain

alex-deaconu:
rvTokenStore persist the RV token via cookie
got rid of userStateReady service (not used)


alex-deaconu:
@olegrise split out functionality from the userstate service. Please review. Thanks.

olegrise:
:+1:

PR #510, opened 2015-04-15
[fix] Save auth_token in localstorage

alex-deaconu:
Fix _state.userToken issues

- Do not pass path on redirect
- Instead save in _state and persist

Simplify url state logic
Remove gapi.auth.setToken calls
Set cookie path for localhost domain


alex-deaconu:
Verified login/logout scenarios with different URLs in Displays and Store.

src/js/svc-rv-token-store.js
```diff
@@ -15,7 +15,9 @@
       var _writeRvToken = function (value) {
         var baseDomain = getBaseDomain();
         if (baseDomain === "localhost") {
-          cookieStore.put("rv-token", value);
+          cookieStore.put("rv-token", value, {
+            path: "/"
+          });
```
olegrise: @alex-deaconu did you add this for consistency sake or something was not working?

alex-deaconu: @olegrise I was seeing duplicate cookies being set while running the Store. Each with different paths. Hence, remove was not necessarily working. 

So, logging in from one URL and logging out from another URL, meant the cookie would remain, and the user could end up being logged in again by switching tabs.

Thanks.

olegrise:
:+1: I tested this with the Store - no issues.

src/js/svc-rv-token-store.js
olegrise: @alex-deaconu not sure I follow your explanation because your fix applies to the localhost only
`if (baseDomain === "localhost") {`. In any case, I'm okay with this change.

alex-deaconu: @olegrise yes, you're right, I forgot to mention this only happens while running Store at `localhost`. Thanks.

alex-deaconu:
@olegrise thanks for review & taking the extra time to verify!

PR #511, opened 2015-04-16
[fix] Added some delay for UX actions

alex-deaconu:
Fixed css selector warnings

Logging out and back in from `shopping-cart-scenarios.js` was failing because of gapi-mock issue. 

However, tests were also randomly failing at different points. I believe it was the action of opening dialogs that was causing this failure. Namely, the transition delay shown when opening the dialog may have caused selectors to come up empty. Added `sleep()` timeouts to alleviate this, so far so good (3 builds in a row, will keep trying). 


olegrise:
:+1: Thanks for fixing this @alex-deaconu !

PR #512, opened 2015-04-17
[fix] Remove sub-company banner when user logs out

alex-deaconu:
When signing out while in a sub-company, the banner remains. Making sure it's removed.


olegrise:
Interesting :+1:

alex-deaconu:
@olegrise this currently happens in the RVA or Displays app. Thanks.

olegrise:
@alex-deaconu I understand. At first it looked strange that `getSelectedCompanyId` is not reset when user logs out because I'm pretty sure it was before, but then I realized that it makes sense.

PR #514, opened 2015-04-23
[fix] Updating province code

alex-deaconu:


brianloosbrock:
@olegrise Instead of changing the Nunavut province code in Store should we just update to this version of CH to use this fix? Thanks

olegrise:
:+1:

olegrise:
@brianloosbrock Good idea. I'll stage it first to make sure there are no surprises.

olegrise:
@brianloosbrock Store that uses new CH is staged at https://store-stage-1.risevision.com

PR #515, opened 2015-04-23
[fix] Ensure name updates when switching between sub-companies

alex-deaconu:
Watching `isSubCompanySelected()` won't trigger the watch if the user switches between sub-companies (flag will be `true` both times). So, the name won't update. Fixing by watching `getSelectedCompanyId()`.


olegrise:
:+1: Right!

PR #518, opened 2015-04-24
Broadcast event when company settings saved

olegrise:
@alex-deaconu what do you think of [this change](https://github.com/Rise-Vision/common-header/compare/feature%2Femit-event-when-company-settings-saved?expand=1#diff-b3d12ec30f34bcaa6dcf8d5cfc6f7e05R63)? The reason I want it is because in the Store Shopping Cart I need to monitor when ShipTo and BillTo addresses change. Address can be changed on the Shopping Cart itself, on Company settings, and when selected company changes. I already have too many $watch functions that already cause Shopping Cart to recalculate taxes (make API calls) several times on page load. I'm thinking having a dedicated event for CompanySettings.Save is much better approach that me trying to add another $watch that will need to watch the change in the every single address field. What do you think?


alex-deaconu:
@olegrise couple of things:
- Your local dependencies, and dependencies.js are out of date, can you do a bower update and build again?
- I'd rather have that event go in the companyState.updateCompanySettings() function, for consistency.
- In the current Common Header improvements doc I'm proposing we add an event for similar purposes: `risevision.company.updated`. Can you take a quick read of the description there and see if it is similar to what you're doing here?

Thanks!

olegrise:
Deleting this PR. Required functionality will be implemented in CH according to the Common Header improvements doc.

Issue #517, opened 2015-04-24
When logging into Store using CH 1.2.2, it just sits there on a blank page after you log in. Refreshing the page shows you the store.


robb-price:
<b>How it works now:</b>
Log into https://store-stage-1.risevision.com/ on Firefox, and notice that once you log in, it just sits on a blank page. If you refresh the page, it works as expected. This does not occur at the production version of the store (https://store.risevision.com)

The same thing does not happen in Chrome, or at http://1-08-008.rvauser2.appspot.com/

ByronDarlison:
@robb-price @alex-deaconu I believe this issue is closed. Can you guys confirm. Thanks

alex-deaconu:
@ByronDarlison @robb-price I'm not aware of this issue still occurring. It may be have been related to a specific, older CH version. @robb-price can you confirm? Thanks.

ByronDarlison:
@alex-deaconu thanks. Closing. @robb-price if you think different let us know...

robb-price:
@ByronDarlison @alex-deaconu Fine with me.

PR #519, opened 2015-04-24
[fix] Clear path if using html5 state

alex-deaconu:
Situation (for Store):
Google authentication redirects to:
`store.risevision.com/#state=...`
Angular HTML5 pushstate replaces location to:
`store.risevision.com/state=...`
CH interprets `state` either way, via `$location.path()`. However, it only clears the hash (`#`) parameter.

This change identifies HTML5 pushstate settings, and clears the `path()` instead of the `#`.


alex-deaconu:
This fixes issue #517 on Firefox. However, it should still be investigated in Store, since it may indicate an underlying problem.

olegrise:
:+1: @alex-deaconu thanks for looking into this.

rodrigopavezi:
@alex-deaconu nice one!!

PR #524, opened 2015-04-29
Removed risevision.user.userSignedIn and

rodrigopavezi:
risevision.user.authorized events
Added risevision.user.signedIn and
risevision.company.selectedCompanyChanged events

@alex-deaconu @olegrise could you please review?

@olegrise we have decided to remove the risevision.user.userSignedIn event which is being used by Store in [here](https://github.com/Rise-Vision/store/blob/c279e0fc4e87ca2b2f7d379489ebb911f0a4a64e/web/script/app.js#L303). I think you can change it to use the risevision.user.signedIn event instead. Moreover I would check whether it is fine that store app reloads each time user logs in.

Cheers


alex-deaconu:
@rodrigopavezi looks good. However we need unit test coverage for all CH events, including `risevision.user.signedOut` and `risevision.company.updated`. Thanks.

rodrigopavezi:
@alex-deaconu good point. I will added them. cheers.

rodrigopavezi:
@alex-deaconu @olegrise please review. Thanks!!

src/js/svc-userstate.js
```diff
@@ -370,6 +365,7 @@
               .then(function (authResult) {
                 if (authResult && !authResult.error) {
                   authenticateDeferred.resolve();
+                  $rootScope.$broadcast("risevision.user.signedIn");
```
olegrise: @rodrigopavezi `if (forceAuth)` is missing. The purpose of this even (at least that's what Store needs it for) is to let application know that user authentication was the result of a user action.

olegrise:
@rodrigopavezi re. Store state reload when sign in event is received - yes it is needed, but thanks for bringing this up.

src/js/svc-userstate.js
rodrigopavezi: @olegrise good point. I think the "risevision.user.signedIn" event should be broadcasted even though the authentication was not triggered by the user. In the case of Store, I think we should have other event to say that the user attempted a authentication. @alex-deaconu what do you think? cheers

olegrise: @rodrigopavezi Store does not need an even for attempted authentication. There is no use case for that. 
To clarify on why Store needs "risevision.user.signedIn" event. There is this situation when Store is run in inRVA mode and in this case authentication happens with popup, not with page redirect. With popup authentication user stays on the same Store page, however if authentication was successful, the page has to be reloaded.

alex-deaconu: @olegrise understood re: store use case. My question is the following: do we need an event for non-user triggered authentication (ie. when the app starts up)? Thanks.

olegrise: @alex-deaconu no, Store does not need it.

rodrigopavezi: @alex-deaconu good point. I don't know about other apps but if we don't need it I think we can restrict the scope of this event to only user actions for now. @olegrise FYI.

rodrigopavezi: @alex-deaconu @olegrise shall I add the if (forceAuth) then? cheers

alex-deaconu: @rodrigopavezi discussed this in more detail with @olegrise and he suggested we add the `forceAuth` parameter in the event `$broadcast` call. This way the Store can verify if this was a User initiated action. Let me know your thoughts on this. Thanks.

rodrigopavezi: @alex-deaconu I think we have to consider if we would like to have this event also broadcast when the app calls the authentication(non-user action). In this case I would suggest to have another event to say that a user made the call. Thus, store could check booth events. 
In case we don't care to listen to this event when app makes the call. We could add the forceAuth check but to not confuse people we would have to state very well this on the documentation of the CH API. 

What do you think?

Cheers

alex-deaconu: @rodrigopavezi ok. If you feel that's the best approach, then please modify the doc with your proposed change to the API.

However, please note, the Store calls the authenticate promise, and instead of using the promise result, it expects an event. 

Thanks.

rodrigopavezi: @alex-deaconu what about if we add forceAuth now and if we see that there is any other case we then revisit it? cheers

alex-deaconu: @olegrise thoughts? Should we get together and have a quick chat about this? Thanks.

olegrise: @rodrigopavezi works for me.

rodrigopavezi:
@alex-deaconu @olegrise I have made the changes as discussed today. Please review. Thanks a mill.

olegrise:
@rodrigopavezi you `dist/js/common-header.js` is out of sync with source. Please re-build.

rodrigopavezi:
@olegrise thanks, good catch. I missed that.

alex-deaconu:
:+1:

rodrigopavezi:
@alex-deaconu @olegrise I moved both back to the authorise function. Should I leave like that or move the userSignin to authenticate? cheers

alex-deaconu:
@rodrigopavezi should be good for now. You have to re-build, and re-base from master. Thanks.

rodrigopavezi:
@alex-deaconu rebase and build done! cheers

alex-deaconu:
Looks good. Thanks!

PR #525, opened 2015-04-29
[chore] Remove bootstrap dependency

alex-deaconu:
angular-bootstrap is used instead

**Note:** this may affect calling apps that use dependencies.js. However, using bootstrap (non angular) is incorrect and should be fixed.


rodrigopavezi:
:+1:

olegrise:
:+1:

PR #527, opened 2015-04-30
[chore] Removed credit cards/payment methods modals

alex-deaconu:
button was hidden; functionality currently exists in the Store

@olegrise @rodrigopavezi please review. Thanks.


rodrigopavezi:
:+1:

olegrise:
:+1: This is something we started working and then left unfinished. There is a chance we still may need to come back to this, but I'm sure requirements will change but that time. I'm okay with deleting this.

alex-deaconu:
@olegrise understood. The code/ux looked out of date so if we do decide to centralize credit card/payment methods in the CH we should import the updated version from the Store. Thanks!

Issue #531, opened 2015-05-11
`selectedCompanyChanged` event not firing on `resetCompany` call


alex-deaconu:
The `risevision.company.selectedCompanyChanged` event is not firing on `resetCompany()` call:
https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-companystate.js#L96

Also unit test coverage is erroneous for this event:
https://github.com/Rise-Vision/common-header/blob/master/test/unit/svc-companystate-spec.js#L60
since the `broadcastSpy` is added here, first event trigger will always fulfil the `calledWith` dependency, so whether or not subsequent event calls trigger the spy is irrelevant.

ByronDarlison:
@alex-deaconu I don't follow this defect. Can you confirm it's status?

alex-deaconu:
@ByronDarlison this was fixed. Closing. Thanks.

PR #534, opened 2015-05-13
[fix] Updated event broadcast logic

alex-deaconu:
Event now called only when a company is switched or reset
This should result in exactly only 1 event per call

Updated unit tests to reflect this and test company reset


alex-deaconu:
Fix for #531.

@tejohnso please review. @olegrise @rodrigopavezi FYI. Thanks!

tejohnso:
:+1:

Issue #538, opened 2015-05-20
CH resets selected company after I open User Settings and then click Save


alex-deaconu:

ByronDarlison:
@alex-deaconu is this still open? I think this is now fixed. Thanks

alex-deaconu:
@ByronDarlison you're right. Closing. Thanks.

PR #539, opened 2015-05-21
[fix] Make sure init doesn't reset selected company

alex-deaconu:
Fixing #538.


olegrise:
Looks good.

Here is something I think that can be improved down the road. Have you noticed that saving user settings makes 4 API calls? Calling just one API method `user.patch` should be enough. Instead CH calls: `user.patch`, `user.get`, `user.get`, `company.get`.

alex-deaconu:
@olegrise looks good. Added https://github.com/Rise-Vision/common-header/issues/542 for this. Thanks.

PR #541, opened 2015-05-21
Broadcast "risevision.user.updated" event when user settings are updated

olegrise:
@alex-deaconu what do you think of this change? Store needs to know when user settings are updated so it can re-run validation on shopping cart page.


olegrise:
@rodrigopavezi, sorry forgot to add you to this PR. Please review.

rodrigopavezi:
@olegrise I think it is good to have it. The only thing missing I guess is the tests. cheers

olegrise:
@rodrigopavezi the problem with unit tests for this fix is that the entire User Settings modal is missing unit tests. I created a separate Trello card for addressing this. Merging this PR.

alex-deaconu:
@olegrise personally, I don't agree with the event being broadcast from a controller. I think it should be broadcast by `userService`. So, this should be changed to be broadcast by `refreshProfile()` once it is complete. Thoughts? @rodrigopavezi FYI. Thanks.

rodrigopavezi:
@alex-deaconu interesting, that is a good point. So you mean that it is better to propagate an event from the service as the controller is specific to a view, right? cheers

olegrise:
@alex-deaconu I knew that you would want it in the service. My problem is that I need an event indicating that user settings have changed. I don't need an event that tells me user settings are simply reloaded.

alex-deaconu:
@rodrigopavezi correct. `refreshProfile()` may be called from elsewhere hence the event will not be triggered. Thanks.

alex-deaconu:
@olegrise okay, I don't disagree. In that case why not add a new `updateUserSettings()` function similar to what we have in `companystate`, and deprecated or make `refreshProfile()` private? Thoughts?

olegrise:
@alex-deaconu might be not a bad idea.

olegrise:
BTW, adding `updateUserSettings()` will also fix issue https://github.com/Rise-Vision/store/issues/1017

PR #549, opened 2015-06-05
[chore] Removed newsletter setting from user settings

alex-deaconu:
They can subscribe when they register
They can unsubscribe from the newsletter itself

@olegrise @rodrigopavezi as discussed, moved this change to its own PR. Please review. Thanks!


olegrise:
@alex-deaconu it looks like you need to update corresponding e2e test.

alex-deaconu:
@olegrise good call. Totally missed that! Fixed and tests are running OK now. Thanks!

olegrise:
:+1:

PR #551, opened 2015-06-12
[chore] Use variable key; check for key

alex-deaconu:
New common-style and angular versions


alex-deaconu:
Yesterday an issue started appearing with unit tests failing in the Displays App; it seems the cause of the issue is loading the Segment tracking code. This PR enables a calling app to modify/reset the Segment API Key. Also, if the key is reset the tracking script will no longer be loaded.

Also added a singleton initialization for the service.

olegrise:
:+1: good to know.

rodrigopavezi:
LG :+1:

PR #554, opened 2015-06-16
[fix] Search box component build results

alex-deaconu:
@olegrise this should fix the issue you're seeing in the Store.

Please review. Thanks!


olegrise:
:+1:

PR #555, opened 2015-06-25
Add `url` property to segment.track()

olegrise:
We need a way to tell if `track` event was triggered on production or other environment.

@alex-deaconu please review. @rodrigopavezi FYI.


alex-deaconu:
@olegrise looks good :+1: . 

Wondering if you can update the unit tests to reflect this change? Thanks!

olegrise:
@alex-deaconu I can't figure out how to test it. I was trying to use `analytics.on('track',...)` as described [here](https://segment.com/docs/libraries/analytics.js/#emitter) - it's not working, then I was trying to use a [callback property](https://segment.com/docs/libraries/analytics.js/#track) - same story - callback is not fired. My next best idea is to add a `_track` wrapper method to the `analyticsEvents`. What do you think?

alex-deaconu:
@olegrise interesting. I'm wondering if you can mock the `$window.analytics` object and go from there. Thanks.

olegrise:
@alex-deaconu I tried that. Same story.

olegrise:
@alex-deaconu interesting idea. That might work. Thanks for the suggestions.

olegrise:
@alex-deaconu added test - see the last commit. Thanks.

alex-deaconu:
@olegrise build failing because of jshint error: analytics is not defined. 

Otherwise, test looks good. Thanks for adding.

olegrise:
@alex-deaconu oh-oh. Fixed it. Sorry about that.

alex-deaconu:
Nice, I like how simple this turned out. Great job!

rodrigopavezi:
@olegrise awesome!!

PR #556, opened 2015-07-06
Enable Intercom Messaging

olegrise:
- enable Intercom Messaging per app manually. Using the trick described [here] (http://docs.intercom.io/configuring-Intercom/in-app-messenger#can-i-make-the-intercom-messenger-only-available-to-selected-users)
- Intercom Messaging can be enabled/disabled via "ENABLE_INTERCOM_MESSAGING" flag in the environment configuration. Ideally it should be done using attribute of the CH directive, but the problem is that `intercomSettings` should be set before `analytics.js` is loaded and that is done way before directive is initialized.

@alex-deaconu @rodrigopavezi please review and confirm if you are okay with the approach.


rodrigopavezi:
LG :+1:

olegrise:
Thanks @rodrigopavezi. I'd like to get @alex-deaconu feedback too, but he is on vacation and I don't think it worth waiting 1 week. I'm merging this PR.

alex-deaconu:
@olegrise looks good. Thanks for looping me in on this.

PR #557, opened 2015-07-07
Fix analyticsEvents not being initialized.

olegrise:
This fixes an issue introduced in PR https://github.com/Rise-Vision/common-header/pull/556. The issue was that "analyticsEvents" dependency was removed from [this line](https://github.com/Rise-Vision/common-header/pull/556/files#diff-2ec175c12c8011573203566f05849457L1435) and as a result analyticsEvents factroy was never created and then `_identify` function was never called.

@rodrigopavezi please review.


rodrigopavezi:
LG :+1:

PR #560, opened 2015-07-15
Fixing link redirect with preventDefault

alex-deaconu:
Looks like I had a workaround in place, but it just wasn't working in this case.

@rodrigopavezi please review, @olegrise FYI. Thanks!


rodrigopavezi:
LG :+1:

olegrise:
Saw it. Thanks.

PR #565, opened 2015-08-07
broken link

ghislaineguerin:


alex-deaconu:
@ghislaineguerin you forgot to build. Fixed. Looks good but let's wait for build to pass. Thanks!

ghislaineguerin:
@alex-deaconu why are the tests failing? is it something that I had to do when changing the url?

alex-deaconu:
@ghislaineguerin no, nothing to do with that. CH e2e tests always tend to fail. Will merge regardless. Thanks.

PR #567, opened 2015-08-14
Fix browser navigation history

olegrise:
@alex-deaconu please review. See explanation in the code comments. This is not ideal solutions, but I tried few things and this is the only one that works. Let me know if you have any questions about this fix.


olegrise:
Oh-oh. Looks like I broke some unit tests. Looking into it now.

alex-deaconu:
@olegrise so I'm assuming just changing the event to `$locationChangeSuccess` did not give the correct results? 

If you've tested and everything looks good, I think we can go with this :+1: . Thanks.

src/js/svc-selected-company-url-handler.js
```diff
@@ -21,10 +21,43 @@
           newCompanyId !== userState.getSelectedCompanyId()) {
           userState.switchCompany(newCompanyId);
         } else if (!newCompanyId && userState.getSelectedCompanyId()) {
+          var currentURL = $location.absUrl();
           $location.search("cid", userState.getSelectedCompanyId());
-          $location.replace();
+          if (currentURL === $location.destUrl) {
+            // see explanation below
+            $location.replace();
```
alex-deaconu: @olegrise shouldn't you be clearing `$location.destUrl`? Or does it get cleared automatically the next time `locationChangeSuccess` is called? Thanks.

olegrise:
@alex-deaconu re. just use `$locationChangeSuccess` event - I tried that and it worked fine with `ng-href`, `ui-sref`, and `stage.go`, but there was some strange inconsistent behaviour when I tried to switch companies. If switching companies worked, then using just `$locationChangeSuccess` event would be the best solution.

alex-deaconu:
@olegrise got it. Thanks for the clarification.

olegrise:
@alex-deaconu Fixed unit test and squashed. Some e2e tests are failing but I assume I can ignore it. Right?

olegrise:
It's green. Merging.

alex-deaconu:
:+1:

PR #569, opened 2015-09-03
Multiple login issue fix

alex-deaconu:
Added token to the gapi object.

Save username instead of token
Authenticate via username if available

New gapi-mock version fixes unit tests

Build results


src/js/svc-userstate.js
```diff
@@ -37,15 +37,14 @@
 
   .value("GOOGLE_OAUTH2_URL", "https://accounts.google.com/o/oauth2/auth")
 
-  .run(["$location", "$window", "userState", "$log",
-    function ($location, $window, userState, $log) {
+  .run(["$location", "$window", "userState", "gapiLoader", "$log",
+    function ($location, $window, userState, gapiLoader, $log) {
```
rodrigopavezi: @alex-deaconu I think I added the gapiLoader here as I was trying to use it here. I believe it is not necessary anymore. Thoughts? cheers

alex-deaconu: @rodrigopavezi removed. Thanks.

alex-deaconu:
To note, we're clearing the `_state.params` variable; we only want to set the token the first time the API gets called; after that the username parameter is populated and used to identify the user.

Otherwise, all authenticate calls rely on the `opts.authuser: username` parameter to select the right account.

This is available for review at:
https://editor-stage-2.risevision.com
https://store-stage-2.risevision.com

@rodrigopavezi please review. @ezequielc @olegrise FYI. Thanks.

rodrigopavezi:
@alex-deaconu I have tried on editor-stage-2 and it seems the login loop issue still there. I cleaned my cookies and cache. I will look further on the code when I got a chance. cheers

alex-deaconu:
@rodrigopavezi looks like stage-2 was overwritten by another branch. Redeployed to https://editor-stage-3.risevision.com. Thanks.

rodrigopavezi:
@alex-deaconu LG, editor-stage-2 is working but not store. Thanks for fixing it. Hopefully we get less users with login problems. :+1:

src/js/svc-userstate.js
```diff
@@ -212,6 +211,13 @@
 
         gapiLoader()
           .then(function (gApi) {
+            // Setting the gapi token with the chosen user token. This is a fix for the multiple account issue.
+            if (_state.params) {
+              gApi.auth.setToken(_state.params);
```
olegrise: @alex-deaconu have you tested if [auto-refresh](https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-userstate.js#L163) still works? We tried using `gApi.auth.setToken` in the past, but the problem was that if you call it manually, then it keeps your token even if `gApi.auth.authorize` returns the new refreshed token. The only way to fix this is to call `gApi.auth.setToken(null)`, but then if user is logged in with multiple accounts, you'll lose the selected account.

alex-deaconu: @olegrise good question. I did test auth-refresh but I don't think I did with that particular use case. I will make the suggested correction (`setToken(null)`) and test to ensure everything works as expected. Thanks.

alex-deaconu: @olegrise I've just finished testing that use case, and as you expected Authentication expires after 1 hour. However, using `setToken(null)` fixes the issue. Will update the code and re-deploy staging. Thanks.

alex-deaconu:
Once Auth is successful, remove the token from memory and call `setToken(null)`.

@olegrise please review the latest commit. Thanks.

olegrise:
@alex-deaconu sounds good. Couple of clarification questions:
1) What happens if user selects second account and then refreshes the page? Will it stay on the second account or will it switch to the first?
2) Re. `setToken(null)` fix - similar to above, will it remember that user was logged in with second account?

alex-deaconu:
@olegrise good questions. We only rely on the Token to *identify* what account the user has selected. Once an account is selected and the user is authenticated, we set `_state.userToken = username` and persist it to local storage.

The key from here (once the Auth token is removed from the picture) is this line (https://github.com/Rise-Vision/common-header/pull/569/files#diff-ee7f135533a1266b5fd6eee4cb5ef4b5R203):
`opts.authuser = _state.userToken;`

Setting the `authuser` parameter forces the API to authenticate that particular username. This is how we *remember* which user was selected once the app is reloaded. 

Thanks.

olegrise:
@alex-deaconu Aaaaaah! Awesome re. setting `opts.authuser`. 
Could you clarify where in the code if you call `setToken(null)`?

alex-deaconu:
@olegrise I wasn't calling `setToken(null)` specifically, just clearing the token and relying on further authenticate API calls to implicitly call it. However, I don't think that's a good approach, and I've updated the code to explicitly call it. See latest commit, again. Thanks!

olegrise:
@alex-deaconu makes sense. I'm going to test this with the Store tonight and then we can merge.

olegrise:
@alex-deaconu I just tried that on https://store-stage-3.risevision.com/ and I can't even login.

alex-deaconu:
@olegrise the last commit I put in had an issue. By setting authToken(null) the oAuth API was failing (there's a Javascript error indicating that).

I've reverted such that the `setToken(null)` happens implicitly in a follow up `authenticate` call. Basically what I had before the latest commit.

Re-building Store right now. Sorry about that.

Thanks.

olegrise:
@alex-deaconu all use cases work like a charm. You found an awesome solution. Buen trabajo ;-)
:+1:

PR #570, opened 2015-09-10
Cookies disabled warning

ezequielc:
@alex-deaconu please review. Thanks


alex-deaconu:
@ezequielc looks good. Can you also remove this service and references to it: 
https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-localstorage.js
It doesn't seem to be used anywhere - since there's a similar service provided by AngularJS. 

Thanks.

ezequielc:
Sure @alex-deaconu , I will remove it.

ezequielc:
@alex-deaconu It is being used by https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-userstate.js

alex-deaconu:
@ezequielc yes it is. And by a couple of other services. But like I said, `localStorageService` is an Angular service, so our own implementation of it never runs. Can you double check that's true and see if our version can be removed? Thanks.

ezequielc:
@alex-deaconu I removed the references from userState and common-header directive. I didn't find any other references. Please review. Thanks!

alex-deaconu:
@ezequielc looks good. @olegrise any concerns with the above? Thanks.

olegrise:
@alex-deaconu @ezequielc looks good. I double checked the Store code - I can't find any references to the `localStorageService`.

PR #571, opened 2015-09-21
Allow application to overwrite closeIFrame() function.

olegrise:
@alex-deaconu what do you think of this change? 

The use case is following:
- User goes to the Store from RVA by pressing "Add from Template" button
- User buys a template
- User close Store by pressing [X] button
- Store return Template ID to the RVA and then RVA creates a presentation

I was thinking of defining a variable which would define action type for the [X] button and then product code that has to be returned, but then realized that probably the cleanest way is to simply let the Store to overwrite the closeIFrame() function.

### DIFF
--- a/src/js/controllers/ctr-close-frame.js
+++ b/src/js/controllers/ctr-close-frame.js
@@ -1,11 +1,15 @@
 angular.module("risevision.common.header")
 
 .controller("CloseFrameButtonCtrl", [
-  "$scope", "$log", "gadgetsService",
-  function ($scope, $log, gadgetsService) {
+  "$scope", "$log", "gadgetsService", "$rootScope",
+  function ($scope, $log, gadgetsService, $rootScope) {
     $scope.closeIFrame = function () {
-      $log.debug("gadgetsService.closeIFrame");
-      gadgetsService.closeIFrame();
+      if (typeof $rootScope.closeIFrame === "function") {
+        $rootScope.closeIFrame();
+      } else {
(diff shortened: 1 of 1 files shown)


alex-deaconu:
@olegrise I like that. I had no idea this functionality even existed.. I wonder if it is used anywhere else.

Do you mind adding unit tests for this controller? Also, you need to add the build results to this PR. 

Thanks.

olegrise:
@alex-deaconu oh, cool. I'll add tests and dist. Just wanted to see what you think of this idea. Thanks!

alex-deaconu:
@olegrise one thing. To be honest I'm not a huge fan of using $rootScope for this override. After investigating further I've found that Angular has ways to override functions:
- Method 1: order of dependency injection can override a service/controller, etc: http://stackoverflow.com/questions/26464347/overriding-angularjs-service
- Method 2: using a decorator or delegate to override all or some functionality:
http://www.mrvdot.com/all/using-decorators-in-angular
http://odetocode.com/blogs/scott/archive/2013/06/06/decorating-angularjs-services.aspx

I'll leave it up to you as to whether or not you want to update this PR as per the above. I was just curious on the subject and wanted to investigate further.

Thanks!

olegrise:
Thanks @alex-deaconu. Interesting ideas. I realized that I can simply overwrite gadgetsService.closeIFrame function in the Store code and won't even need to make any changes to the CH. I think I'll take that route because the only app we need this change for is Store.

PR #579, opened 2015-10-08
Update e2e tests to use common-e2e

alex-deaconu:
Removed mocks from e2e index page
Updated scenarios to use real data
Updated to use common-e2e components

Various template fixes based on e2e tests

Update circle.yml


rodrigopavezi:
@alex-deaconu looks good. :+1:  Curious, what do you have in mind about the commented out tests? cheers

alex-deaconu:
@rodrigopavezi I was just going to add some clarifications regarding those.

To note:
- registration tests are currently disabled for 2 reasons: It's being refactored so I'm waiting for the new version, and it requires an extra google account we can use specifically for registration.
- company move scenarios are disabled: complexity, I will try to add those scenarios back in later
- shopping cart, last 2 scenarios are disabled: We're adding fake products to the cart, and the store server returns 503 error and doesn't persist them. Need to fix this either by mocking that specific API or adding real products.
- system messages scenarios mostly disabled: No system messages are available to test with from the Core. Need to fix by using a mock.

All the above are fixable but require more work. I want to merge this PR first before dealing with those specific issues.

Thanks.

rodrigopavezi:
@alex-deaconu sounds good! good job! cheers

PR #584, opened 2015-10-13
Unify tests into one file

alex-deaconu:
Renamed to homepage to be consistent with other apps
Created common-scenarios which runs all 'logged in' tests

Remainder of tests need to be integrated into common-scenarios,
but currently they require the user to be logged out

Added verification that secondary (jenkins1) credential is present
Trying to fix issue where jenkins company could be deleted


alex-deaconu:
One more iteration for e2e test updates. At this point, I think this is the final one unless any other test failures occur.

@rodrigopavezi @ezequielc please review. Thanks.

rodrigopavezi:
@alex-deaconu looks good. :+1:

PR #586, opened 2015-10-15
Add remaining scenarios to test-suite

alex-deaconu:
Add sign-out function
Improvements to test-suite.scenarios.js
Add account removal at the end

Use only one username, register account, test, CH and delete itself


alex-deaconu:
As discussed, only using 1 user for these tests. The only caveat is that 2 builds can't run at the same time because they'll be competing for that user. 

The alternative is to keep the same structure as before and use 2 users.

@rodrigopavezi @ezequielc thoughts?

Thanks.

rodrigopavezi:
@alex-deaconu looks good. :+1:

PR #587, opened 2015-10-16
Do not show cookie error if server connection fails

alex-deaconu:
Updating css & dist


alex-deaconu:
Tested both server failures (by sending the http request to an invalid URL) and w/ Third Party cookies disabled option.

@ezequielc @rodrigopavezi please review. Thanks!

ezequielc:
@alex-deaconu Did you check the case were the URL is under a firewall?

alex-deaconu:
@ezequielc no, what is the easiest way to do that? Also, what should the result be? Thanks.

ezequielc:
@alex-deaconu I am not sure how we could test it. Ideally, we should show a specific warning about firewall, but I think you can just display the cookies warning and in the docs we have references to enable cookies and to check firewall. Thoughts?

alex-deaconu:
@ezequielc I would imagine behind a firewall the `$http` request would fail, so it would not show the warning (as the code currently stands). However, I tend to disagree, showing the cookie warning in that instance may be confusing to the user. We would need to enable a separate, firewall detection if we wanted to capture that scenario. Thoughts? Thanks.

ezequielc:
@alex-deaconu I agree, I also think the best option is to have another message for that. I guess we should talk with the team about it.

alex-deaconu:
@ezequielc agreed. I will bring this point up in the card, and circle everybody in. Thanks.

ezequielc:
@alex-deaconu  :+1:

PR #591, opened 2015-10-21
Fix/mobile modal form

RV-Adrian:
@ezequielc 
I've built the dist. Please see my question below, thx!


dist/js/common-header.js
```diff
@@ -5588,34 +5598,6 @@ angular.module("risevision.common.header")
 
 })(angular);
 
-/*
```
RV-Adrian: @ezequielc 
Can't see a reason why this was removed?

ezequielc: @RV-Adrian It should be there. I run your branch locally and it didn't work because of the missing config.
I run `gulp build` and it was added back. Maybe you are missing the config file?

rodrigopavezi: @RV-Adrian @ezequielc I looked at your previous commit. I think you didn't run the gulp build so it did not add the changes to the common-header.js file. cheers

ezequielc:
@RV-Adrian :+1:

dist/js/common-header.js
RV-Adrian: @rodrigopavezi @ezequielc I did run 'gulp build' but it was out of the usual sequence so.. something went off.  Thanks for the assistance!

PR #595, opened 2015-10-28
Fix forms for mobile

ezequielc:
- remove stop-event directive from CH
  - use stop-event on modals
@rodrigopavezi Please review. Cheers


rodrigopavezi:
@ezequielc looks good :+1: cheers

PR #598, opened 2015-11-12
Require role directive from Editor App

alex-deaconu:
Include coveralls configuration

Moving directory here so it can be used in other Apps.


ezequielc:
@alex-deaconu :+1:

PR #602, opened 2015-12-07
Remove the empty state from dropdowns

alex-deaconu:
When a user clicks on the field, the empty state option will not show

Added directive which will automatically transform the 'null' value to '' (blank)

Fixed unit tests for directives by excluding partials
Fixed failing test by removing properties


alex-deaconu:
Couple of things on this PR: 
Because of the file pattern for unit tests here: https://github.com/Rise-Vision/common-header/blob/master/gulpfile.js#L68, tests for dtv-require-role.js were not running. Fixed by modifying file name, but this caused an issue with i18n strings. To fix this, I've created a separate module for directives in the CH.

Also, the fix for the issue on hand is caused by this: https://groups.google.com/forum/#!topic/angular/Q8PzDmJ8pGs. Basically, despite the fact that the value of the dropdown option was set to `""` (blank), it would default to `null` causing Core errors. Fixed this by adding a new directive as suggested in the issue.

However, also hiding the empty option from the list via `ng-hide="true"` so the user will only have valid options in the select list.

olegrise:
Interesting...

alex-deaconu:
Staged at https://displays-stage-0.risevision.com.

alex-deaconu:
@olegrise I would imagine the same issue may exist with Store Address fields. Whether or not the Store server has issues accepting `null` values I don't know, but you should at least consider putting in the `ng-hide="true"` to prevent the user from accidentally selecting the `blank` value. Thanks.

src/templates/company-fields.html
```diff
@@ -40,8 +40,8 @@
         Country
       </label>
       <select id="company-settings-country" class="form-control selectpicker"
-        ng-model="company.country" ng-options="c.code as c.name for c in countries">
-        <option value="">&lt; Select Country &gt;</option>
+        ng-model="company.country" ng-options="c.code as c.name for c in countries" empty-select-parser>
+        <option ng-show="false" value="">&lt; Select Country &gt;</option>
```
ezequielc: @alex-deaconu Just to confirm, why do we need `ng-show="false"` ? I guess it works like a placeholder, am I right? Thanks

alex-deaconu: @ezequielc exactly. The `empty` option will only show if no value has not been previously selected. However if the user clicks on the input, that option is not available in the list. Thanks.

ezequielc: Answered in your previous comment. Thanks

ezequielc:
@alex-deaconu Looks good. Nice work! :+1:

alex-deaconu:
@ezequielc thanks!

olegrise:
Thanks @alex-deaconu. 
@RV-Adrian see above and let me know if you think we need to make the same change in the Store UI. Basically the change here is that `< Select Country >` is excluded from the dropdown list so a user cannot choose that option. The only way a user can choose `< Select Country >` is if they don't choose anything because `< Select Country >` is the default option :wink: .

RV-Adrian:
@olegrise Gotcha! I think this would make a lot of sense, yeah.
(Address editor, payment form, and elsewhere maybe..)
Surprised this one got so complex, but glad a solution is there.
Who should set up this task card?

PR #605, opened 2016-01-15
Redirect to original request after signing in

ezequielc:
@alex-deaconu Please review. Thanks


src/js/svc-userstate.js
```diff
@@ -319,11 +319,22 @@
...
-            loc = $window.location.origin + "/";
-            // Remove first character (/) from path since we're adding it to loc
-            path = $window.location.pathname ? $window.location.pathname.substring(
-              1) : "";
-            search = $window.location.search;
+            var redirectTo = localStorageService.get(
+              "risevision.common.redirectTo");
+            if (redirectTo) {
+              var parser = document.createElement("a");
+              parser.href = redirectTo;
```
olegrise: @ezequielc do you think using `$document[0]` instead of `document` would be a better option?

ezequielc:
@olegrise @alex-deaconu I am closing this PR as I found a simpler solution that changes only apps repo. Thanks!

PR #606, opened 2016-01-20
Update navigation

ezequielc:
@alex-deaconu Please review.

I had to manually set jshint version as the new version was breaking the build.
There is an open issue in their repo for that.

Staged at: https://apps-stage-4.risevision.com/

Thanks


alex-deaconu:
@ezequielc looks good :+1: .

Will this jshint issue affect other repos as well? Thanks.

ezequielc:
@alex-deaconu I guess it will affect, whenever we rebuild without cache. Unless they fix it before that. 
If you get a jshint fail with "xxx was used before it was defined" message, just use the version "2.8.0".
Reference: https://github.com/jshint/jshint/issues/2833
Thanks!

PR #607, opened 2016-01-21
Update ui-router 'cid' parameter if available

alex-deaconu:
Added ui-router as dependency


alex-deaconu:
This PR is part of the task for opening links in Apps in a new window. The main obstacle to doing that was that URLs generated by the `ui-router` did not contain the `cid` parameter.

After a bit of investigating, I have figured out how to integrate the `cid` parameter from the CH with `ui-router` in apps. Declaring the parameter as part of the route achieves this (https://github.com/Rise-Vision/rise-vision-apps/compare/feature/link-url?expand=1#diff-8ec8983b9f4c33db4ddae15c48d179b5R99). 

This is great because now all `ui-sref` and `$state.go` links will automatically carry the `cid` with them. No need to add it back in every time. Only `ng-href` links that miss the `cid` will need to be updated. So, scenario #2 described in the comments (https://github.com/Rise-Vision/common-header/pull/607/files#diff-2ec175c12c8011573203566f05849457R5726) is avoided.

**Note:**  this PR doesn't modify any of the old code. The feature is only enabled if `cid` exists in `$stateParams` (which it doesn't if not declared in the path).

olegrise:
@alex-deaconu nice!

alex-deaconu:
Unfortunately the above implementation does not give the correct results in some situations. The main issue is that when the `cid` parameter is updated, the whole route refreshes, reloading the `resolve` dependencies, and causing double API calls in some situations.

Due to ui-router issues, this route refresh cannot be prevented in the `$state.go()` call.

So, I've updated the implementation to use a simpler solution, which seems to work correctly in our situations, however still uses the `$location.search()` function: https://github.com/Rise-Vision/common-header/pull/608. 

Closing this PR. Thanks.

PR #608, opened 2016-01-28
Update ui-router 'cid' parameter if available

alex-deaconu:
Added ui-router as dependency


dist/js/common-header.js
```diff
@@ -1620,9 +1621,7 @@ angular.module("risevision.common.header", [
 
 .run(["$rootScope", "userState", "selectedCompanyUrlHandler",
   function ($rootScope, userState, selectedCompanyUrlHandler) {
-    $rootScope.$watch(function () {
-        return userState.getSelectedCompanyId();
-      },
+    $rootScope.$on("risevision.company.selectedCompanyChanged",
```
alex-deaconu: Changing the event on which the `cid` parameter is updated in the URL. This event fires "quicker" than the $watch. Prevents issue where watching same event in the Apps lists causes `cid` to never update.

dist/js/common-header.js
```diff
@@ -5660,24 +5659,37 @@ angular.module("risevision.common.header.directives")
...
+    "$location", "userState",
+    function ($state, $stateParams, $location, userState) {
+      // Called when the selectedCompanyId is changed
       this.updateUrl = function () {
         var selectedCompanyId = userState.getSelectedCompanyId();
         // This parameter is only appended to the url if the user is logged in
+        // Do not apply during $state.trasition (handler will)
         if (selectedCompanyId && $location.search().cid !==
-          selectedCompanyId) {
+          selectedCompanyId && !$state.transition) {
```
alex-deaconu: Do **not** update params during `$state.transition`; it causes $state reload.

Let them be updated after the transition by the handler via `updateSelectedCompanyFromUrl`.

dist/js/common-header.js
```diff
@@ -5660,24 +5659,37 @@ angular.module("risevision.common.header.directives")
...
+          $stateParams.cid = selectedCompanyId;
+          $state.params.cid = selectedCompanyId;
```
alex-deaconu: "Secretly" update the parameters. By pre-populating the UI router parameters with the `cid` value it won't force a `$state` reload when the URL is updated.

alex-deaconu:
Added inline descriptions for the changes; added to `dist` by mistake.

To summarize, this is a follow up to the previous CH PR for this. This much simpler version simply updates the `cid` parameters in the current `$state`. This is done "secretly" as in without notifying ui-router of the change, to prevent the `$state` reload.

This change relies on the `cid` parameter being added to the `ui-router` path. Otherwise, that parameter is (should be) simply ignored.

To verify:
- Loading links without the `cid` should not cause double API calls, and also should update the selected company
- `ui-sref` links should have the `cid` as long as it is defined in the path
- Switching companies should work as expected; lists should refresh, and manage pages should stay the same (not load the API again)
- If a user accesses a path while logged out, that path should be persisted. Once logged in they should be directed to the page they requested
... (will add more usecases if needed)

Staged at https://apps-stage-4.risevision.com.

@olegrise @ezequielc please review. Thanks!

ezequielc:
@alex-deaconu Looks good! :+1: Thanks!

alex-deaconu:
@olegrise I'm wondering if you have any concerns regarding this branch and the Store. 

I've staged this CH version at https://store-stage-2.risevision.com/ and from my testing there's nothing out of the ordinary. @ezequielc FYI.

Just want to validate this and then I will merge.

Thanks!

olegrise:
@alex-deaconu no concerns. From the Store point of view everything appear to be working as expected. Code changes look good as well. Thanks.

PR #609, opened 2016-02-03
Fix Issue #46: Modal for timeline doesn't close when clicking outside

ezequielc:
@alex-deaconu Please review. Thanks!


alex-deaconu:
@ezequielc I don't believe the Timeline modal is loaded from the CH. It's just particular to Apps: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/index.html#L73

Thanks.

ezequielc:
@alex-deaconu I will try again locally. It didn't work until I updated CH, maybe it was cache. I'll let you know. Thanks

ezequielc:
@alex-deaconu I will close this PR as is it not needed indeed. Thanks for catching it.

PR #610, opened 2016-02-04
Cache authorize promise to prevent concurrent calls

alex-deaconu:
If _authorize is called more than once concurrently, an API failure is bound to happen
Prevent this by caching the promise while in use

Fix failing e2e tests


alex-deaconu:
Currently there's an issue with loading the CH in a separate tab. 

To replicate, go to apps.risevision.com/editor and open several Presentations (by cmd+click) in separate tabs. If you do not go to those tabs for a couple of seconds, once you do select one you should see the login screen (rather than the Editor). The App does not recover unless you reload. Demo: http://screencast.com/t/3JvrQgJw.

After further investigation, it seems the oAuth API is failing if more than one call is made concurrently. We have a "visibility" handler which causes the second call (https://github.com/Rise-Vision/common-header/blob/master/src/js/svc-userstate.js#L146) when switching to the new tab.

Fixed by preventing the call from calling the API twice (caching the promise).

Also fixed e2e tests which were giving problems. Instead of clicking the links and opening in new tabs, we're just checking the href & target parameters for those buttons.

alex-deaconu:
Staged at https://apps-stage-4.risevision.com/.

olegrise:
@alex-deaconu :+1: . Good one!

PR #613, opened 2016-02-19
Hide Rise roles for non Rise users

olegrise:
@ezequielc please review


src/js/svc-userstate.js
```diff
@@ -510,10 +510,10 @@
           return _state.inRVAFrame;
         },
         isRiseAdmin: function () {
-          return hasRole("sa");
+          return hasRole("sa") && companyState.isRootCompany;
```
ezequielc: @olegrise Is `companyState.isRootCompany` intentionally missing parenthesis? Thanks

ezequielc:
@olegrise Just one question, otherwise looks good. Cheers

olegrise:
@ezequielc good catch. I'm surprised my tests worked. Fixed parenthesis and squashed.

ezequielc:
@olegrise Looks good! :+1: Thanks!

olegrise:
@ezequielc added missing `!` in the `isRootCompany` function. The root company should not have a parent!

ezequielc:
@olegrise Oops, I missed that too. Isn't the field named `parentId`? If that is the case, it might be blank not only null. Thanks

olegrise:
@ezequielc OMG, it's embarrassing. Of course it's `parentId`. Fixed. Thanks.

ezequielc:
@olegrise :) No worries. 
I think you should also check for blank/empty string in `parentId`. I just verified and I got a blank value when using my Rise account. Sorry if I wasn't clear in my previous comment. 
Thanks

olegrise:
@ezequielc if `parentId` is empty string or undefined or null, won't this `!_state.userCompany.parentId` return TRUE?

ezequielc:
@olegrise Sorry, you are right, empty string will work properly. Looks good :+1: . Thanks!

olegrise:
Thanks @ezequielc !

olegrise:
Oops, forgot to squash.

olegrise:
@ezequielc this is tagged as `v1.9.3`

ezequielc:
Thanks @olegrise

PR #614, opened 2016-03-07
Improvements do BQ logging

ezequielc:
@alex-deaconu  A few things in this PR:
- Updated externalLogger to use Template Tables (templateSuffix)
- Logging to a new BQ table due to schema changes
- Added bigQueryLogging which auto-populates companyId and userId to BQ events. I was going to add it as a new method to externalLogging but due to circular references with userState I created a separate service.
- Added a flag to only log from Production.
- Logging "User Registered" to BQ.

Please review.  Thanks!


alex-deaconu:
@ezequielc looks good :+1: . 

Not necessary since the code hasn't really changed, but it would be nice to have a test for the `templateSuffix` field. 

Thanks.

ezequielc:
@alex-deaconu Good point. I will try to add a test for it. Thanks!

ezequielc:
@alex-deaconu I added a test for `templateSuffix`. There is a nice `useFakeTimers()` method in sinon. Please review. Thanks!

alex-deaconu:
@ezequielc looks good :+1: . 

Nice re: `useFakeTimers`, cool feature for this particular test.

Thanks!

PR #628, opened 2016-05-05
Fix #434 - add user cache

### LINKED ISSUE #434: When you mark a user as inactive and create a new user with the same name in the same Company, when you click on the user, it brings up the old inactive user instead of the new active one
<b>How it works now:</b>
1. Mark a user as inactive in your Company.
2. Create a new user with the same username and email. 
3. Click on the newly created user in the Company Users list, notice that it brings up the old INACTIVE user, instead of the new one.

For further clarification: https://dl.dropboxusercontent.com/u/8237394/Screencasts/2014-12-19_1543.swf

ezequielc:
- Apply changes from Rise-Vision/ng-core-api-client#27

@alex-deaconu Please review. Thanks!


alex-deaconu:
👍

PR #639, opened 2016-11-03
Add call to identify user/company after accepting terms

ezequielc:
- added call to identify() after accepting terms
- updated common-app-components dependency

@fjvallarino Please review. Thanks!


fjvallarino:
:+1:

PR #641, opened 2016-12-01
Reorganize menu items

fjvallarino:
This adds the Alerts and Store Account links to the Company menu


ezequielc:
@fjvallarino The icons for User Settings and alerts are swapped. Thanks

src/templates/auth-buttons-menu.html
```diff
@@ -5,12 +5,26 @@
...
-    <a href="" ng-click="userSettings()" class="user-settings-button action">
+    <a href="" ng-click="alertSettings()" class="alert-settings-button action">
       <i class="fa fa-user"></i>
+      <span class="item-name">Alert Settings</span>
+    </a>
+  </li>
+
+  <li ng-show="isRiseVisionUser">
+    <a href="" ng-click="userSettings()" class="user-settings-button action">
+      <i class="fa fa-bullhorn"></i>
```
fjvallarino: @alex-deaconu for the Store Account one?

alex-deaconu: @fjvallarino I just realized I put the comment in the wrong spot. I was suggesting to use actual hyperlinks for both "Alert Settings" and "Store Account". Thanks.

fjvallarino: @alex-deaconu I missed it for the Store Account one (it was supposed to open in a new tab). Do you think it makes sense for the Alerts one too? I think it's better to keep it on the same tab, at least when you're running inside Apps.

alex-deaconu: @fjvallarino I think same tab for Alerts is fine. Thanks.

fjvallarino:
@ezequielc @alex-deaconu updated

Should I look around to add an unrelated e2e test case to fix this Coveralls issue?

alex-deaconu:
@fjvallarino I've updated the setting so those notifications are no longer sent. 0.1% drop in test coverage should be fine. Thanks.

alex-deaconu:
@fjvallarino looks good 👍 . Thanks for fixing.

PR #643, opened 2016-12-09
Update angular to 1.5.9; fix common-i18n issue

alex-deaconu:


alex-deaconu:
Using Angular 1.6.0 is throwing a lot of errors in the Unit Tests: `Possibly unhandled rejection`
https://circleci.com/gh/Rise-Vision/common-header/1331

Using Angular 1.5.9 instead.

Also updated the common-i18n version and fixed some dependencies.

fjvallarino:
@alex-deaconu one option is explicitely ignoring that error: https://github.com/angular-ui/ui-router/issues/2889#issuecomment-257654153

alex-deaconu:
@fjvallarino yes I saw that reply as well. Tried it out:
```
app.config(['$qProvider', function ($qProvider) {
    $qProvider.errorOnUnhandledRejections(false);
}]);
```
and it resolves the unit test errors and the errors shown in Apps. There is one Apps unit test that fails related to the i18n library, which can most likely be fixed.

However, my thinking is that since Angular 1.6.0 is quite fresh, and it is causing some backwards compatibility issues, maybe we can hold off upgrading as of yet.

@ezequielc FYI.

Thoughts? Thanks.

fjvallarino:
@alex-deaconu agreed

alex-deaconu:
One further fix for this PR. After 1.4 the `ngModel.$parsers` behaviour has changed slightly. Due to this, the `integer-parser` directive was no longer working correctly. Fixed by applying the recommended implementation from the documentation: https://docs.angularjs.org/api/ng/directive/select
(see bottom section: "Binding select to a non-string value via ngModel parsing / formatting") with a few tweaks.

The Directive is used in the Timeline component, Display Settings (Timezone field) and Company Settings (Timezone field).

ezequielc:
@alex-deaconu 👍

alex-deaconu:
@olegrise FYI this is the update I was talking about at SoS. Thanks.

olegrise:
Thank you @alex-deaconu. I created a card on the Store board to upgrade to the 1.5.9.

alex-deaconu:
@olegrise sounds good. Merging this for now; we can resolve any Store specific issues when working on that card. Thanks!

olegrise:
@alex-deaconu sounds good.

PR #646, opened 2017-01-12
Fix/container testing

RV-Adrian:
@ezequielc @alex-deaconu @fjvallarino 
Release tag PR as discussed. Thanks!


alex-deaconu:
@RV-Adrian my suggestion is that when creating a PR for the common-header, you don't include the dependencies.js/dependencies.min.js files in the commit. In this case you're inadvertently updating the Angular js version, which could cause undesired effects. 

Also, the common-header build includes common-app-components dist files. Those are missing in this PR. Thanks.

RV-Adrian:
@alex-deaconu 
Dependencies reverted. 'gulp build' didn't create any diffed common-header*.js files it seems.

alex-deaconu:
@RV-Adrian interesting, I guess those files are not included in the build. Since this PR only updates the common-style version, I would suggest we just close it. We can update the version tag in the following PR you have open. Thanks!

RV-Adrian:
@alex-deaconu Agreed. Thx!

PR #648, opened 2017-01-23
Class Name adjust for Padding utility

RV-Adrian:


RV-Adrian:
@ezequielc @alex-deaconu @fjvallarino 
See above.  Going to tag you directly on PRs here, rather than Trello, going forward. Hopefully that is easier to organize. Thanks!

ezequielc:
@RV-Adrian Changes look good. 
Could you add your common-style branch to `bower.json` so we can verify it doesn't break any test? Also, you missed `gulp bump` to update  the version. Thanks!

ezequielc:
@RV-Adrian Thanks for updating. 👍 Before merging, don't forget to update common-style with the new tag. Cheers!

RV-Adrian:
@ezequielc Will do! 'gulp bump' is new to me, I'll keep that in mind next time. 👍

alex-deaconu:
@RV-Adrian this PR updates the Angular JS version in the `dependencies.js` file to `1.5.11`. I don't think that's intentional and should probably be removed. Thanks.

RV-Adrian:
@alex-deaconu I was wondering if that would be an issue. How do I avoid that using 'gulp build'?

alex-deaconu:
@RV-Adrian it's not really an issue, as the dependencies update automatically, but I am just suggesting we minimize the risk of this PR. 

In terms of `gulp build`, you can't really prevent it. Just don't include dependencies.js/dependencies.min.js in your commits.

Thanks.

RV-Adrian:
@alex-deaconu Got it thanks! Updated now.

alex-deaconu:
@RV-Adrian 👍 . Thanks for fixing!

bower.json
```diff
@@ -49,6 +49,7 @@
     "angular": "~1.5.0",
     "angular-mocks": "^1.5.5",
     "angular-sanitize": "~1.5.0",
-    "angular-bootstrap": "~0.13.0"
+    "angular-bootstrap": "~0.13.0",
+    "rv-common-style": "fix/padding-utility"
```
alex-deaconu: @RV-Adrian this is a branch name - should reference a release tag here instead. Thanks!

bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "rise-vision-common-header",
-  "version": "1.11.9",
+  "version": "1.12.0",
```
alex-deaconu: @RV-Adrian this version number does not match the one for package.json. Thanks.

alex-deaconu:
@RV-Adrian I don't think this PR was ready for merge. Please see inline comments. Thanks!

bower.json
RV-Adrian: @alex-deaconu I think I saw the thumbs up and jumped the gun. :) Reverting on a branch now.

bower.json
RV-Adrian: @alex-deaconu I'll update this as well. Does 'gulp bump' usually tackle both files?

alex-deaconu: @RV-Adrian yes, `gulp bump` will increment the minor versions of both files. By minor versions I mean the third digit - `1.11.*10*` in this case.

In fact, as a general rule we don't increment the second digit unless we implement major changes; nothing we can do about it now since it was merged but going forward try to increment the last number only.

Thanks!

PR #650, opened 2017-01-24
Fix/padding utility

RV-Adrian:


RV-Adrian:
@alex-deaconu @ezequielc @fjvallarino 
I had to revert this one and fix a bower resolution + package.json version number.
Please validate, thanks!

alex-deaconu:
@RV-Adrian my suggestion is to set the version to `1.12.1` since `1.12.0` will need to be replaced. Thanks.

alex-deaconu:
@RV-Adrian you missed one file to bump. Fixed with another commit, so we should be good to go. 👍

RV-Adrian:
@alex-deaconu Thanks! Deploying now 👍

PR #654, opened 2017-02-09
Fix angular-mocks dependency

alex-deaconu:
Didn't increment versions, as there is no code change. Just fixing build, fix similar to the one in Apps.


fjvallarino:
:+1:

PR #658, opened 2017-02-16
Run e2e tests off test Core

alex-deaconu:
Load config.js after userstate.js to overwrite CORE_URL value
Add e2e test block for using test core


alex-deaconu:
I did not spend too much time on the test; and it probably shows (to note none of the index.html elements referenced in other tests are referenced from `e2e/pages`). But, it serves it's purpose.

fjvallarino:
:+1:

PR #661, opened 2017-02-22
fix on trial and trial expired cases

settinghead:


settinghead:
@alex-deaconu @olegrise turns out there are a couple of cases missing for subscription status (i.e. `on-trial` and `trial-expired`). Here is the updated logic. Could you review again? (Sorry about the many PR's).

olegrise:
@settinghead makes sense 👍

ezequielc:
@settinghead Looks good 👍 Thanks

PR #662, opened 2017-02-24
Fix Store URL config issue

alex-deaconu:
Fixes issue with Storage subscription URL (and maybe other URLs that use that parameter).

Staged at https://apps-stage-0.risevision.com/


ezequielc:
@alex-deaconu 👍

PR #663, opened 2017-03-09
Show a message-box with instructions for duplicate user

alex-deaconu:


alex-deaconu:
Changed from showing an Alert to a MessageBox. Updated User Add and Settings controllers; using common-i18n due to the strings being duplicated in both. See separate i18n PR.

I'm not really sure why coverage dropped so dramatically due to these changes.

src/js/controllers/ctr-user-add-modal.js
```diff
@@ -50,7 +51,19 @@ angular.module("risevision.common.header")
...
-            alert("Error" + humanReadableError(error));
+
+            var errorMessage = "Error: " + humanReadableError(error);
+            if (error.code === 409) {
+              var errorMessage1 = $filter("translate")(
+                "common-header.user.error.duplicate-user-1");
+              var errorMessage2 = $filter("translate")(
+                "common-header.user.error.duplicate-user-2");
+
+              errorMessage = errorMessage1 + $scope.user.username +
```
ezequielc: @alex-deaconu Instead of appending, I am wondering if we could use variable replacement (https://angular-translate.github.io/docs/#/guide/06_variable-replacement). Thoughts? Thanks

alex-deaconu: @ezequielc good call. Updated code, also fixed a small issue with strings. Opened an i18n PR for this change & updating apps staging once that build is complete. Thanks.

ezequielc:
@alex-deaconu Thanks 👍

PR #665, opened 2017-03-10
Send additional information to Zendesk widget

settinghead:
Some temporary DOM manipulation is added to alter the form content on ZD Widget.

Related updates:
- https://github.com/Rise-Vision/rise-vision-apps/pull/361
- https://github.com/Rise-Vision/store/pull/1248


settinghead:
Apps: @ezequielc @alex-deaconu @fjvallarino 
Delivery: @stulees  @tejohnso @rodrigopavezi 

Could one of you from each team review this update? Thanks!

dist/js/common-header.js
```diff
@@ -4075,14 +4076,67 @@ angular.module("risevision.common.support", [
...
+
+      function _activate() {
+        var username = userState.getUsername();
+        var identity = {
+          email: username,
+        };
+
+        $window.zE.identify(identity);
+
+        _startDomMonitor();
```
ezequielc: @settinghead Could you clarify why we need this monitor? Is it to update username and company id in the form? If that is the case, could we watch the company change event? Thanks

settinghead: @ezequielc the monitor looks for an open Zendesk widget on the page (every 1 second). If it finds one, it will proceed to fill in some fields of the form, and hide them (i.e. set `display` to `none`). That said, it would not work by subscribing to company change event or any of the existing events in CH. The reason I'm doing interval polling is because there unfortunately isn't an event you can subscribe with ZD widget API.

We tried different solutions, although I understand this is essentially hacking, it's by far the only correctly working solution to send additional information along with a ZD ticket.

ezequielc:
@settinghead Thanks for the clarification. Code looks good. Do you have a staged version I could test? Thanks

ezequielc:
@settinghead I found a small issue. If we open Zendesk widget and close it before entering the form, the timer doesn't stop.
If we open it a second time, we have two timers running and if we change the route, only one timer gets canceled.
Is there a way to detect widget being closed so we can clear each related timer? 
Thanks

settinghead:
@ezequielc 

Jut pushed an update to resolve the duplicate timer issue. 

re. Cancelling timer on widget close: I tried to look into the DOM structure but wasn't able to find a good way to detect if a widget is closed (once initialized, the ZD widget actually never closes but only becomes invisible). 

I think it's OK to keep the timer running at 1 second interval even when the widget is closed because first, it's not that frequent and second, the timer will clear itself when the user navigates away from the current route.

ezequielc:
@settinghead Sounds good. Thanks! 👍

PR #667, opened 2017-03-28
Honor supscription if cancelled but not expired

settinghead:


settinghead:
Apps:
@alex-deaconu  or @ezequielc please review.

Delivery:
@tejohnso @rodrigopavezi or @stulees please review.

Thanks!

alex-deaconu:
@settinghead do you have a Company ID I can test this with? Thanks.

alex-deaconu:
@settinghead curious, why not just re-use this service instead of implementing a custom one: https://github.com/Rise-Vision/component-subscription-status/blob/master/src/svc-subscription-status.js ? Thanks.

settinghead:
@alex-deaconu I think that logic is for `on-trial` and is no longer relevant here. Did you mean to update `component-subscription-status` on that part?

alex-deaconu:
@settinghead yes, and yes. 

To confirm, I think the `expiry` field was just recently added for `Cancelled` subscriptions. In the subscription-status-component we check a second API to get the same information. 

The component is already pre-loaded in the CH. If you updated the code in the component to do the same thing, we could knock 2 birds with 1 stone (re-use the component code and fix the issue with the current subscription-status).

Not sure re: timelines, and complexity, so I'll leave it up to you to decide whether to pursue or not.

Thanks!

settinghead:
@alex-deaconu I took a look at `component-subscription-status` and it seems there is part of the logic that I don't yet fully understand. I'll have to leave it as-is for now because I need to push CH update by EOD today. Maybe will take a look back later. Thanks for the suggestion, though!

alex-deaconu:
@settinghead sounds good 👍 . Thanks!

olegrise:
@settinghead one thing I would be careful about is comparing expiry date returned from server with the date on user's computer. The PC date can be changed to anything. So it the worst that could happen is that they get a free support, then I guess it's not a big deal, but still not desirable. The alternative, if I'm not missing anything, is to use [authorization API](https://docs.google.com/document/d/148NxsVDU1nfQgX5a_2VHXguO7LJhj12FHPmCNItNBRc/edit?usp=sharing) instead of subscription status API. The authorization API returns `“authorized”:true` when subscription is cancelled before the subscription end date.

settinghead:
@olegrise ah just merged. You have a good point on handling date checking on client side only. Let me take a look at the Authorization API.

alex-deaconu:
@olegrise great point, thanks for bringing this up. @settinghead I don't think it's necessary to revert this PR as the code is not deployed. Thanks.

settinghead:
@olegrise @alex-deaconu I thought about it, and came to realize that the current implementation is inherently unsound from an authorization's standpoint, doesn't matter if we use subscription status API or authorization API. Both will be subject to potential user meddling (if they choose to do so).

The key problem here is subscription status is passed to the ZD system _by client JavaScript_ as the single source of truth. So even if we use Authorization API, the user can still bypass the superficial roadblock in browser by directly opening Zendesk widget via JavaScript manipulation. 

We talked about this before among Delivery, and the conclusion was that from a business's standpoint, such imperfect solution is acceptable for now, until evidence shows otherwise. There might be some determined users who want to hack his way to get support, but we estimated the chance of that happening is small and even if it happens, it's manageable. 

(BTW In order to have a fully sound implementation, I believe something must to be done at the server level. One possibility via the Zendesk API Proxy I recently set up. )

Let me know if the above makes sense and if you are OK with the solution. Thanks!

olegrise:
@settinghead good with me. Thanks.

alex-deaconu:
@settinghead 👍

PR #674, opened 2017-04-19
Fix/store link

RV-Adrian:


RV-Adrian:
@alex-deaconu 
Looks like the build number was snagged, so I did another bump.  Sound ok?

alex-deaconu:
@RV-Adrian sounds good 👍

PR #676, opened 2017-04-21
common-app-components updates

alex-deaconu:
Deprecated common-svg repo


alex-deaconu:

fjvallarino:
:+1:

alex-deaconu:
@olegrise FYI with this update we are deprecating the https://github.com/Rise-Vision/common-svg repository. What this means is that the Store will no longer require that dependency: https://github.com/Rise-Vision/store/blob/master/web/index.html#L73 .

The svg functionality will be available directly from common-header.js.

Thanks.

PR #680, opened 2017-05-05
Add website field to company settings and registration modals

olegrise:
@alex-deaconu could you please review?


src/js/controllers/ctr-registration-modal.js
```diff
@@ -77,13 +87,14 @@ angular.module("risevision.common.header")
 
           action.then(
             function () {
-              userState.refreshProfile().then()
+              userState.refreshProfile()
+                .then(updateCompanyWebsite)
```
alex-deaconu: @olegrise wondering if we can do this asynchronously. ie. Just call updateCompanyWebsite in the background and not make the user wait for the response. Thanks.

src/js/controllers/ctr-registration-modal.js
```diff
@@ -68,7 +78,7 @@ angular.module("risevision.common.header")
 
 
           var action;
-          if (newUser) {
+          if ($scope.newUser) {
             action = registerAccount(userState.getUsername(), $scope.profile);
```
alex-deaconu: @olegrise you may have already checked, but is there no way to `registerAccount` and create the User's company while adding the website field in one call? Thanks.

src/templates/registration-modal.html
```diff
@@ -43,6 +43,14 @@ <h2 class="modal-title">Let's finish with your details</h2>
           <p ng-show="forms.registrationForm.email.$invalid && !forms.registrationForm.email.$pristine"
             class="help-block validation-error-message-email">Enter a valid email.</p>
         </div>
+        <!-- Website -->
+        <div class="form-group" ng-show="newUser">
+          <label for="website">Company Website</label>
```
alex-deaconu: @olegrise wondering if we should just call this `Website`, for consistency. Thanks.

alex-deaconu:
@olegrise PR looks good. Couple of questions/suggestions. Thanks..

src/js/controllers/ctr-registration-modal.js
olegrise: @alex-deaconu interesting point. You think something like this would be a good solution?
```
              userState.refreshProfile()
                .then()
                .finally(function () {
                  updateCompanyWebsite();
                  analyticsEvents.identify();
```

alex-deaconu: @olegrise I think so. Should speed things up. Also, if updateCompanyWebsite fails for some reason, the User would still be shown registration success. Thanks!

src/js/controllers/ctr-registration-modal.js
olegrise: @alex-deaconu I did check and also discusses the API change with @alexey-rise . The conclusion is that we don't want to add a `website` field to the `account.add` API call. This is consistent with how we update User fields. We call the `account.add` first and then `user.patch`.

alex-deaconu: @olegrise sounds good, thanks!

src/templates/registration-modal.html
olegrise: @alex-deaconu I asked Brian. He thinks (and I agree with him) that it's a good idea to call it `Company Website` for better clarity.

alex-deaconu: @olegrise sounds good. Thanks!

olegrise:
@alex-deaconu please review the last commit.

alex-deaconu:
@olegrise looks good, thanks for fixing 👍 !

olegrise:
@alex-deaconu thanks for reviewing 😃 !

PR #683, opened 2017-05-16
Fix linkCid directive

olegrise:
@alex-deaconu this is the fix for cid attachment not working for Apps links in the Store's navigation menu. What I found is that when `userState.authDeferred.promise` is used inside `state` `resolve` [here for example](https://github.com/Rise-Vision/store/blob/master/web/script/app.js#L491), then [this line](https://github.com/Rise-Vision/common-header/blob/master/src/js/directives/dtv-link-cid.js#L23) is never executed because even though the `$scope.$watch()` is triggered, but the condition `newValue !== oldValue` is always false. It seams like the initial `$scope.$watch()` when `oldValue` is not set yet is somehow lost. I figured that having a local variable inside `linkCid` directive would make the logic more reliable, but if you have better ideas let's discuss.


alex-deaconu:
@olegrise to confirm, this fixes the issue in the Store? Is it currently staged?

In Apps, we have a service that looks like this: https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/common/services/svc-access.js for verifying access. Do you think you would like to re-use that in the Store (ie. we can move that service to CH). 

Thanks.

olegrise:
@alex-deaconu staged at https://store-stage-1.risevision.com

Re. `canAccessApps` service - it's not exactly what I need. In fact the `canAccessApps` service calls `userState.authenticate(false)` which returns `authenticateDeferred.promise`. I'm not sure that using `canAccessApps` instead of `userState.authDeferred.promise` will solve the problem. Regardless, I don't think `canAccessApps` is a good fit for us.

alex-deaconu:
@olegrise looks good 👍 . Thanks!

PR #688, opened 2017-07-17
Fix the way to detect iframes due to recent change in Zendesk Widget

settinghead:


alex-deaconu:
@settinghead looks good, however, don't forget to bump the CH version. Thanks!

settinghead:
@alex-deaconu ah good reminder. Thanks!

PR #698, opened 2017-08-15
ICP Data Collection implementation

alex-deaconu:
Adding test coverage reports for watch


alex-deaconu:
Staged at: https://apps-stage-3.risevision.com/

ezequielc:
@alex-deaconu Do you have a company ID we could use to test this? Thanks

alex-deaconu:
@ezequielc unfortunately, no. By design the ICP modal pops up for a specific user that signed up more than 2 weeks ago and less than 2 months ago.

However, I've staged a version of the PR that removes the date restrictions (https://apps-stage-4.risevision.com) you can use to test the modal itself. 

Let me know if this helps. @fjvallarino FYI. Thanks.

ezequielc:
@alex-deaconu I tested and it looks good. 
Two small issues: In Company Settings, Website field is a bit off margin.
Also, when I am looking at a subcompany, the modal appears and the fields are populated with my parent company, which kinda makes sense as my Role should related to the parent company, but it can cause confusion. I am wondering if we should just not show the modal in that case. Thoughts? Thanks!

alex-deaconu:
@ezequielc good points, addressed the two in another commit. Re-staged at https://apps-stage-3.risevision.com/. Thanks!

ezequielc:
@alex-deaconu I reviewed the code and it looks great 👍 
I was thinking about i18n but I saw it is not implemented in the settings modals, so I am okay with leaving as it. Thanks

alex-deaconu:
@ezequielc @fjvallarino I've added a follow up commit after my investigation into the Hubspot field migration. To confirm, the field names will remain the same (they are re-mapped to the correct names in the hubspot-connector repo), however, the values need to match the values defined in Hubspot. Hence, the value mappings for all the fields has been updated. 

Further, the companySize field requires an Integer value (String field with Integer value) so I've updated the values to just have the lower boundary of the size range. Let me know if you'd like me to adjust that.

Finally, the companyIndustry's "Other" field has no mapping in Hubspot. This results in a red mapping with the original value we send: 
screen shot 2017-08-17 at 2 20 49 pm
This can be fixed later by adding a field named "Other" to map with our results.

Please review the final commit. I will try to stage this version, however data from the test server goes to a different Hubspot sandbox environment. Thanks!

ezequielc:
@alex-deaconu 👍

PR #700, opened 2017-08-22
Add Other to User Company Role

alex-deaconu:
Staged at http://1-09-016.rvauser2.appspot.com/.

@ezequielc @fjvallarino @olegrise please review. Thanks!


ezequielc:
@alex-deaconu 👍

olegrise:
@alex-deaconu 👍 

Side note. I don't know if you've noticed that "IT/Network Administrator" option is selected by default visually, but if you close the dropdown then you'll that nothing is selected. It's not a big deal. Just wanted to bring it up in case you didn't notice.
screen shot 2017-08-22 at 4 18 46 pm

alex-deaconu:
@olegrise yes, that is true. Actually, the default option is "" (blank) in the code (https://github.com/Rise-Vision/common-header/blob/master/src/templates/company-icp-modal.html#L38) with the text "<Select Role>". However, this option is hidden from view in the dropdown so that the user is forced to select an option once they click on it. It's a bit misleading I agree, but it is working as intended. Thanks.

alex-deaconu:
@ezequielc @fjvallarino @olegrise just a quick update, added the "Automotive" icon as well that was missing before. @RV-Adrian FYI, thanks for the link!

ezequielc:
@alex-deaconu 👍

PR #704, opened 2017-09-25
Added website validation

alex-deaconu:
Validation for Company Settings & Registration

Fix issue with ICP dropdown for Add Company

Staged at https://apps-stage-1.risevision.com

@ezequielc @fjvallarino @olegrise please review. Thanks!


olegrise:
@alex-deaconu it's quite often that people enter an email address into website field. That creates a problem when we call HibSpot API. We need to make sure our website validator does not allow email addresses.

alex-deaconu:
@olegrise good point. Added additional validation for the `@` character.

For example, something like `email.add@gmail.com` would not be allowed. However, `email.add/@gmail.com` would be, since the slash is a URL identifier. The second example does look like a valid URL so I think that's a good compromise.

Re-staged Apps for verification. Let me know your thoughts. Thanks!

olegrise:
@alex-deaconu I didn't have time to review the code, but staged version works like a charm! Thanks.

ezequielc:
@alex-deaconu One minor issue. If current website is invalid, when I open Company Settings it doesn't allow me to Save, as expected, but the field is not highlighted as invalid, so I am not aware of the issue. Sample company: https://apps-stage-4.risevision.com/?cid=554d3ef9-78b7-4726-8f46-c1ec140c166d
Thanks

alex-deaconu:
@ezequielc hmm, good point. That's a common pattern that we have, checking for both `$error` & `$pristine`. However, I think in this case it's safe to remove the `$pristine` as this is not a required field. Applied this fix, and restaged.

Let me know if it makes sense. Thanks!

ezequielc:
@alex-deaconu Thanks for fixing it. 
I found another issue, sorry :) If I enter "www.risevision.com", the field is valid, no error message as expected, but the Save button is not enabled.
Same issue happens in the Create Account form, I can't submit with the above url. 
If I type "http://www.risevision.com", everything works fine. 
Thanks!

alex-deaconu:
@ezequielc good catch. As discussed, I've updated the fields to use type `text` rather than `url`. This should fix the issue you encountered. Thanks!

ezequielc:
@alex-deaconu Looks good 👍 Thanks for fixing it!

PR #709, opened 2017-10-13
Disable Change Password for any user other than the currently logged in

fjvallarino:
@ezequielc please review


src/templates/user-settings-modal.html
```diff
@@ -45,12 +45,12 @@ <h2 id="user-settings-label" class="modal-title">
...
       </div>
       <div class="col-xs-6 text-right">
-        <span ng-if="!isAdd">
+        <span ng-if="editingYourself && !isAdd">
           <a href="" class="btn btn-default btn-sm change-password" ng-click="toggleChangePassword()">Change password</a>
         </span>
       </div>
     </div>
-    <div id="passwordForm" class="animated" ng-if="!isAdd" ng-show="isRiseAuthUser && showChangePassword">
+    <div id="passwordForm" class="animated" ng-if="editingYourself && !isAdd" ng-show="isRiseAuthUser && showChangePassword">
```
ezequielc: @fjvallarino Do you need `editingYourself` here? If we hide the button above, how could this be displayed? Thanks

fjvallarino: @ezequielc I think that actually applies to line 18 too. I added them for the sake of safety, but removing them probably makes sense

ezequielc: @fjvallarino I am good with keeping them for now. Thanks

ezequielc:
@fjvallarino 👍

PR #712, opened 2017-10-20
Use animations for Password Change dialog

fjvallarino:
@Rise-Vision/apps please review (deployed to stage-4)


alex-deaconu:
@fjvallarino code itself looks good, but the animation for the password fields looks really weird. Not sure if it's an improvement over the previous version. @ezequielc thoughts? Thanks.

ezequielc:
@fjvallarino I agree with @alex-deaconu. The 'up' animation doesn't look great. I am wondering if you can apply 'fadeIn' class instead of 'fadeInUp'. Maybe it will require @RV-Adrian help. Thanks

RV-Adrian:
@fjvallarino @alex-deaconu @ezequielc 
I agree it's jarring. Initially, that was the intent since the Password form fields had 'appeared' without notice when using a standard fade. 
I can look into other animation properties, or add some other visual queue if it can wait for Monday.

ezequielc:
@RV-Adrian Yes, it can wait. We can release without it and update later. Thanks

PR #724, opened 2017-11-01
Option to hide custom auth

alex-deaconu:


alex-deaconu:
Just includes userState changes for RVA.

PR #738, opened 2017-12-05
Hide reset password button for non admins

alex-deaconu:
Save button functions as password reset save
However, save button is hidden, so they can't reset

Hide button temporarily until a better flow is defined

User can just use the logged out password flow (via email)


alex-deaconu:
As discussed, we'll be hiding the "Reset Password" buttons for users who do not have access to Save. Relatively temporary solution until we have a better reset flow.

PR #743, opened 2017-12-14
Check for 403 error before showing registration

alex-deaconu:
Added check in auth factory
If Core refreshProfile responds with 403
Then allow registration state

If Core responds with something else
Then consider the user signed out

Token expired or unauthenticated response is 401
Deals with other error responses (just in case)

Refactored reset functionality
Use cases were different for each scenario
but the end result is the same, so now
it just resets everything

Also modified registration service
This change is not needed, but I added it
just in case


alex-deaconu:

How to replicate: https://www.screencast.com/t/q0Yt2SPwe

The testing I've done on this is:
- Expired token scenario (see video)
- Unregistered user scenario
- Terms not accepted scenario
- Sign in/Sign out for both GA and Custom Auth

PR #744, opened 2017-12-19
Feature/ux plan upgrade

RV-Adrian:


RV-Adrian:
@ghislaineguerin 
Please review the Common Style portion of this pull request (since it's not integrated with Common Header, there will eventually be a 'dev' branch for the CSS but probably easier to review here)

Highlights:
I broke out the logic of the viewport calculations into editor-layout.scss . It's far from ideal and a bit hacky, but made it way easier to deal with the various viewport states. This will be addressed better eventually on this card https://trello.com/c/3u2cDkmG/2354-ux-combine-all-three-playlist-types-to-one-flexible-class-3

There's also a new 'alert' class for the plans (active and suspended)

Thanks!

PR #745, opened 2017-12-20
Show message when using Google account for Custom Auth

fjvallarino:
@alex-deaconu please review


src/js/components/userstate/controllers/ctr-login.js
```diff
@@ -28,16 +28,21 @@ angular.module("risevision.common.components.userstate")
...
             })
-            .then(null, function () {
-              $scope.errors.loginError = true;
+            .catch(function (err) {
+              if (err.status === 400) {
+                $scope.messages.isGoogleAccount = true;
+              } else { // No special case for 404, for security reasons
+                console.error(err);
+                $scope.errors.loginError = true;
+              }
```
alex-deaconu: @fjvallarino unrelated, but it doesn't seem that we show an "Account not confirmed" exception when they try to log in (to catch the Core's ConflictException - 409 I believe). I think that was missed adding this and it may be causing major problems because the Users don't know they haven't confirmed their account. Thoughts? Thanks.

fjvallarino: @alex-deaconu that's a good point. I have the feeling those features were there at some point in the development, mainly because of the resend confirmation email part. Anyway, I'll add them tomorrow morning.

fjvallarino:
@alex-deaconu updated to include Confirmation Email logic

src/js/components/userstate/controllers/ctr-request-confirmation-email.js
```diff
@@ -0,0 +1,44 @@
+"use strict";
+
+angular.module("risevision.common.components.userstate")
+  .controller("RequestConfirmationEmailCtrl", ["$scope", "$loading", "$log",
+    "userauth",
+    function ($scope, $loading, $log, userauth) {
+      $scope.forms = {};
+      $scope.credentials = {};
+      $scope.errors = {};
```
alex-deaconu: @fjvallarino I don't believe this variable is being used. Thanks.

alex-deaconu:
@fjvallarino PR looks good 👍 .

To confirm, since we're making changes to the error codes returned in the Core, I think we need to deploy this new CH version to all the apps (Apps, Store, Developer hub) to ensure Users receive consistent messages.

Thanks!

fjvallarino:
@alex-deaconu updated.

Yes, unfortunately we will need to update all the projects using Common Header.

PR #749, opened 2018-01-03
Feature/hologram style guide

RV-Adrian:
@alex-deaconu @fjvallarino @ezequielc 
Please review this setup for the early structure  of the 'Rise Vision Style Guide'.

This is incorporated into the Gulp build process.  It builds the guide to the ./styleguide' directory, and the instructions for making updates to this guide are included in the MD file there.  If it's not clear how it works from that text, let me know.

A couple of options to consider:
- We can add the './styleguide' to .gitignore to keep the repo cleaner, and leave it up to the user to build locally only?
- We can also give the Style Guide an URL for everyone to access, at some point.

- This uses a themed dependency (hologram-github-theme) but I'd ultimately like to fork this and make our own customizations to the header, footer, interactions etc. Do you see any issues with that?


RV-Adrian:
@alex-deaconu @fjvallarino 
Here's a video walkthrough to point out the key additions and how they'll work. Thanks!
https://drive.google.com/file/d/1wrzbWH6XFmN8WmkbzeewY1J0u4kWexhS/view?usp=sharing

RV-Adrian:
@ghislaineguerin Please see attached text and video above for how it's all set up, and how it should work going forward. Thx!

alex-deaconu:
@RV-Adrian looks good to me. 

One suggestion is maybe don't add the `hologram` task to the `gulp build` process. This way we are not forced to have hologram installed in order to build CH. Thoughts?

Regarding your questions (1 & 2), I did some research and we can try to use Github Pages to host the Style Guide. According to their documentation (https://help.github.com/articles/user-organization-and-project-pages/#project-pages), if you host your HTML in a `/docs` folder in your repository, the page will be available at a URL similar to rise-vision.github.io/common-header. I think that's worth a try.

Otherwise, PR looks good.

Thanks.

RV-Adrian:
@alex-deaconu 
If the style guide is not built via gulp with 'gulp hologram', then it would have to be done with the command line via 'hologram hologram_config.yml'. 
That's ok with me, but may be inaccessible for some others. And the other downside is.. we don't automate the style guide being updated with every build (which is kind of a bonus, no?)

Sounds good on the URL via Github. Let's leave that for a future improvement on this. Thanks!

RV-Adrian:
@alex-deaconu PS I added you as a reviewer

alex-deaconu:
@RV-Adrian actually that's my main problem with this change. It's not visible: I have to download the repo and run a local HTML page. If it's not hosted anywhere it doesn't seem very useful. Long story short, I do think we should try to use the github pages feature from the get-go. It's just a directory name change.

Otherwise, I'm ok with leaving it in the `gulp build` task, not totally convinced that installing a separate dependency (`gem install hologram`) should be a requirement for running `gulp build`.

Curious, have you looked at https://www.npmjs.com/package/node-hologram?

Thanks.

RV-Adrian:
@alex-deaconu 
re: Gem
Yes, exactly what I wanted to try! Most setup instructions just went with the gem package but I'll give this a whirl.  Don't want anyone jumping through hoops.

Agreed, ultimately we want this to be a reference everyone can bookmark. I'll also try to get an URL going along with NPM

alex-deaconu:
@RV-Adrian disregard re: gem. I tried locally and it seems that's a requirement. Don't think we can do anything about that at this point. Thanks.

alex-deaconu:
Updated PR to target the `/docs` folder for the hologram build so we can have this available via GitHub pages. Some cleanup.

RV-Adrian:
@alex-deaconu 👍

PR #751, opened 2018-01-09
Show current plan and plans modal

fjvallarino:
This PR builds on the previously created branch, feature/plan-upgrade, which has changes to Subscription Status service (it seems I did not build the last commit on that branch, since some of those changes show up here as new on the dist folder).
 
A pretty large refactoring was done in https://github.com/Rise-Vision/common-header/commit/4a671c36c3b958be4667261d4288efe088d4800c, so probably reviewing individual commits is not going to be super enlightening.


src/templates/plans-modal.html
```diff
@@ -0,0 +1,40 @@
...
+    <button type="button" class="close" ng-click="dismiss()" aria-hidden="true">
+      <i class="fa fa-times"></i>
+    </button>
+    <h3 class="modal-title">Choose Your Plan</h3>
+  </div>
+  <div class="modal-body u_padding-lg" stop-event="touchend">
+    <div class="grid-list row">
+      <div class="col-xs-12 col-sm-6 col-md-3" ng-repeat="plan in plans">
+        <div class="panel panel-default">
+          <div itemscope="" itemtype="http://data-vocabulary.org/Breadcrumb">
```
alex-deaconu: @fjvallarino curious, what is the purpose of this div? Thanks.

fjvallarino: @alex-deaconu not sure, it was in the prototype and left it there to avoid breaking something. I think @RV-Adrian is the right person for this question.

src/js/controllers/ctr-plans-modal.js
```diff
@@ -0,0 +1,67 @@
...
+
+    $scope.getPlansDetails = function () {
+      $loading.start("plans-modal");
+
+      planFactory.getPlansDetails()
+        .then(function (plans) {
+          $scope.plans = plans;
+        })
+        .catch(function (err) {
+          $log.debug("Failed to load detauls", err);
```
alex-deaconu: @fjvallarino typo here :)

alex-deaconu:
@fjvallarino PR looks good to me 👍 . Thanks!

src/templates/plans-modal.html
RV-Adrian: @fjvallarino @alex-deaconu 
.grid-list styles the .panel children inside it. 
Basically set this way so as to re-use the existing .panel style without breaking the CSS hierarchy as-is.

alex-deaconu: @RV-Adrian I was referring to this: `<div itemscope="" itemtype="http://data-vocabulary.org/Breadcrumb">`. I'm not sure what itemscope and itemtype do on this div, or if they're needed. Thanks.

RV-Adrian: @alex-deaconu @fjvallarino 
I see, it appears that's just some legacy code (initial version had this layout as per a Store list, so that got ported over). 
I'm not sure what it does, but I doubt it's relevant in this popup. Please delete, thanks.

PR #755, opened 2018-01-15
Show Plans modal on related Store links

fjvallarino:
This PR adds:

- Logic to show Plans modal for Storage, Priority Support, Templates Library and Embedded Presentations.
- Moves the Plan related code into a self contained component, to overcome an issue with Storage Selector (otherwise, a dependency to Common Header would be needed)
- Updates the text of Free Plan


src/js/components/plans/svc-plan.js
```diff
@@ -30,14 +27,30 @@
...
+      "subscriptionStatusService", "currencyService", "PLANS_LIST",
+      function ($q, $log, $rootScope, $modal, $templateCache, userState, storeAPILoader, subscriptionStatusService,
+        currencyService, PLANS_LIST) {
         var _factory = {};
         var _plansCodesList = _.map(PLANS_LIST, "pc");
         var _plansByType = _.keyBy(PLANS_LIST, "type");
         var _plansByCode = _.keyBy(PLANS_LIST, "pc");
 
+        $rootScope.$on("risevision.company.selectedCompanyChanged", function () {
+          console.log("SVC", userState.getSelectedCompanyId());
```
alex-deaconu: @fjvallarino this console message need to be cleaned up or updated. Thanks.

src/js/components/subscription-status/dtv-subscription-status.js
```diff
@@ -19,23 +18,21 @@
             expandedFormat: "@",
             showStoreModal: "=?",
             customProductLink: "@",
-            customMessages: "@"
+            customMessages: "@",
+            customOnClick: "&"
```
alex-deaconu: @fjvallarino curious, is this component ever used without a `customOnClick` function? Also, do we use any other `customOnClick` function than `showPlansModal`?

Maybe we can simplify. Thoughts? Thanks.

alex-deaconu:
@fjvallarino reviewed, couple of questions/comments. Thanks.

src/js/components/subscription-status/dtv-subscription-status.js
fjvallarino: @alex-deaconu display-details.html uses it (it was not requested to show Plans at this point), and I'm not sure if other application other than Apps uses it. Originally I had the showPlansModal as part of the subscription-status directive (you can see that in one of the first few commits), but there were two problems:

1. It didn't feel correct to have the logic inside the directive
2. It was causing hell when running the e2e test cases for subscription-status

To be honest, I refactored it because of (2), but I still feel it's the best choice.

alex-deaconu: @fjvallarino sounds good. Thanks.

fjvallarino:
@alex-deaconu added one more commit

alex-deaconu:
@fjvallarino 👍

PR #758, opened 2018-01-17
Add e2e test cases for Plans

fjvallarino:
I can only test conditions for the Free Plan. I will add the other three scenarios (basic, advanced and enterprise) until next epic, when we can switch plans from Apps.


src/templates/components/plans/plans-modal.html
```diff
@@ -18,11 +18,13 @@ <h1>${{plan.priceMonth}}</h1>
...
-                 target="_blank" class="cta_button btn btn-white u_margin-lg">Current Plan</a>
+                 target="_blank"
+                 class="cta_button btn btn-white u_margin-lg current-plan">Current Plan</a>
               <a ng-show="canUpgrade(plan)"
-                 target="_blank" href="https://store.risevision.com/product/{{plan.productId}}" class="cta_button btn btn-primary u_margin-lg">Subscribe</a>
+                 target="_blank" href="https://store.risevision.com/product/{{plan.productId}}"
+                 class="cta_button btn btn-primary u_margin-lg subscribe-plan">Subscribe</a>
               <a ng-show="canDowngrade(plan)" ng-click="showDowngradeModal()"
-                 class="cta_button btn btn-default u_margin-lg">Downgrade</a>
+                 class="cta_button btn btn-default u_margin-lg downgrade-plan">Downgrade</a>
```
alex-deaconu: @fjvallarino wondering why you're opting for css classes rather than ids in this partial. Generally we found that adding css classes for e2e tests is confusing (because we don't know that those are actually valid classes or not). Wondering if you could use ids instead, or annotate the classes with e2e__ or the like. Thanks.

alex-deaconu:
@fjvallarino looks good. I think the simple tests suffice. Just one comment. Thanks.

src/templates/components/plans/plans-modal.html
fjvallarino: @alex-deaconu good point. I'll check if using id works, because the ng-repeat means there will be four elements with the same id. In case it doesn't, I'll add the e2e_ prefix.

src/templates/components/plans/plans-modal.html
```diff
@@ -5,7 +5,7 @@
     </button>
     <h3 class="modal-title">Choose Your Plan</h3>
   </div>
-  <div class="modal-body u_padding-lg" stop-event="touchend">
+  <div class="modal-body u_padding-lg plans-modal" stop-event="touchend">
```
alex-deaconu: @fjvallarino I think you missed a spot ;)

src/templates/components/plans/plans-modal.html
alex-deaconu: @fjvallarino sounds good, thanks for addressing!

src/templates/components/plans/plans-modal.html
fjvallarino: @alex-deaconu oh, I copied that pattern from company-settings-modal. I'll switch it too.

alex-deaconu:
@fjvallarino 👍

PR #764, opened 2018-01-29
Add port to gapi initialization

alex-deaconu:
* Remove cookie policy from signIn
Doesn't seem to be needed

* Move utils to gapi component
Prevents dependency issues with gapi-loader

* Use window.location.port
$location.port() returns 80 & 443 for http and https respectively
Adding those ports breaks some gapi.auth2 functionality


alex-deaconu:
This is deployed at both https://apps-stage-0.risevision.com & https://apps-stage-1.risevision.com to allow testing with different subdomains.

Localhost domain requires port otherwise gapi authentication is not retained.
Custom Auth does not work if the port is applied to that cookie, so the port is only applied to gapi.auth2 init.

@ezequielc @fjvallarino @rudibravo please review. Thanks!

PR #766, opened 2018-02-06
Feature/update ind trials

RV-Adrian:


RV-Adrian:
@ghislaineguerin Please review! 
This is a common-header update to support the 'Remove Individual Product Trials' wireframe, staged at https://apps-stage-4.risevision.com/ 
(select Plans Popup)

image

The class addition is not ideal, but I think works while we are deciding on a better solution to get Plans descriptions into Apps.

RV-Adrian:
@fjvallarino 👍

ghislaineguerin:
@RV-Adrian All looks good, but I think we'll have reusability issues with the pseudo-content. Can we change that?

RV-Adrian:
@ghislaineguerin I was wondering about that as well. When it hits development I will work with devs to see other options via Angular. Thx!

PR #770, opened 2018-02-14
Apply Plan Banner updates

fjvallarino:
This is ready for review on stage-2. We have to discuss tomorrow how we can test the different statuses (I tested locally by hard coding the different scenarios)


fjvallarino:
@Rise-Vision/apps @rudibravo this is the PR we discussed during stand up.

src/js/components/plans/svc-plan.js
```diff
@@ -69,7 +72,8 @@
                   var priceMap = _.keyBy(plan.pricing, "unit");
                   var price = priceMap[monthKey] || {};
 
-                  plan.type = plan.name.toLowerCase().replace(" plan", "");
+                  plan.name = plan.name.replace(" Plan", "");
```
alex-deaconu: @fjvallarino just wondering why we're not doing the 'lowercase' replace here anymore? Don't think it's wrong, just maybe safer to use. Thanks.

fjvallarino: @alex-deaconu mainly because I need to keep the first part of the plan name capitalized (for instance, of "Advaced Plan" I keep "Advanced").

src/js/components/plans/svc-plan.js
```diff
@@ -86,37 +90,6 @@
           return deferred.promise;
         };
 
-        _factory.getCompanyPlan = function (companyId) {
-          $log.debug("getCompanyPlan called.");
-          var deferred = $q.defer();
-
-          subscriptionStatusService.list(_plansCodesList.slice(1), companyId)
```
alex-deaconu: @fjvallarino I believe the `subscriptionStatusService` is no longer needed in this service, and can be removed. Thanks.

alex-deaconu:
@fjvallarino looks good. One question and one comment. Thanks.

src/js/components/plans/svc-plan.js
alex-deaconu: @fjvallarino okay, makes sense. However, you also added the `plan.name` property in the `PLANS_LIST` variable, so maybe this is just not needed in that case. Thanks.

fjvallarino: @alex-deaconu the problem is this list is retrieved from Store, not from our PLANS_LIST constant (we get the short description and price)

alex-deaconu: @fjvallarino missed that. Disregard. Thanks!

alex-deaconu:
@fjvallarino looks good 👍 . Thanks for fixing!

PR #772, opened 2018-02-19
Plan Modal updates to Start Trials

rudibravo:
I've to move the store authorization service from rise-vision-apps to start the trial here.
It's deployed to stage-3 for verification.

Thanks


src/js/components/plans/ctr-plans-modal.js
```diff
@@ -2,7 +2,10 @@ angular.module("risevision.common.components.plans")
 
 .controller("PlansModalCtrl", [
   "$scope", "$modalInstance", "$log", "$modal", "$templateCache", "$loading", "planFactory", "currentPlan",
-  function ($scope, $modalInstance, $log, $modal, $templateCache, $loading, planFactory, currentPlan) {
+  "allPlansMap", "storeAuthorization",
+  function ($scope, $modalInstance, $log, $modal, $templateCache, $loading, planFactory, currentPlan,
+    allPlansMap, storeAuthorization) {
```
fjvallarino: @rudibravo I think we don't need the allPlansMap variable to be provided as a dependency, since we load it when the modal is invoked (line 16 in your version). The map you will receive after that call is keyed by "type", so you will need to update the ```canStartTrial``` function a bit.

The main reasons for this is ```getCompanyPlan``` method was removed from svc-plan.js in my PR (since we now rely on the information Core provides, instead of determining the current plan on the client side)

rudibravo: @fjvallarino good point. I didn't notice the getPlanDetails method already had that map. I'll change it accordingly. Hopefully it will make easier to merge the PR as well. Thanks.

src/js/components/plans/ctr-plans-modal.js
```diff
@@ -58,6 +61,31 @@ angular.module("risevision.common.components.plans")
...
+        allPlansMap[plan.productCode].statusCode === "trial-available") {
+
+        return true;
+      }
+
+      return false;
+    };
+
+    $scope.startTrial = function (plan) {
+      storeAuthorization.startTrial(plan.productCode);
```
fjvallarino: @rudibravo it would be safer to close the modal after the startTrial promise returns (mainly to be able to show an error in case the call fails). Usually we display a spinner when making a remote call; you can see an example in product-details-modal.html/ctr-product-details-modal.js in App, in particular how it uses $loading service and rv-spinner* directives.

rudibravo: @fjvallarino good point.

rudibravo:
@fjvallarino I've addressed the issues you pointed out. Thanks

src/js/components/core-api-client/svc-store-authorization.js
```diff
@@ -0,0 +1,68 @@
...
+            deferred.reject(e);
+          });
+
+        return deferred.promise;
+      };
+
+      return factory;
+    }
+  ])
+  .factory("checkTemplateAccess", ["storeAuthorization", "TEMPLATE_LIBRARY_PRODUCT_CODE",
```
alex-deaconu: @rudibravo see comment re: this factory in the Apps PR. Thanks.

src/js/dtv-common-header.js
```diff
@@ -31,7 +31,8 @@ angular.module("risevision.common.header", [
   "risevision.common.components.svg",
   "risevision.common.components.plans",
   "risevision.common.support",
-  "risevision.common.email"
+  "risevision.common.email",
+  "risevision.common.store.authorization.services"
```
alex-deaconu: @rudibravo wondering why the lengthy module name. Why not something simple like `risevision.store.authorization`, following the pattern of other module names in this component. Also, maybe group this module above with the other modules from that component. Thanks.

src/js/components/plans/ctr-plans-modal.js
```diff
@@ -58,6 +66,47 @@ angular.module("risevision.common.components.plans")
       return false;
     };
 
+    $scope.canStartTrial = function (plan) {
+
+      console.log(plan);
+      console.log($scope.allPlansMap);
```
alex-deaconu: @rudibravo are these $log messages necessary or for debugging? Thanks.

src/js/dtv-common-header.js
rudibravo: @alex-deaconu I tried to follow the convention I saw in the few services I checked. I can rename the module name or move it together with another module. What would be the preferable way? and what module exactly?

alex-deaconu:
@rudibravo reviewed. So far other than the few comments inline, no major issues. However, I'd like to see this PR rebased off the main branch now once Francisco's PR has been merged. Maybe there are some simplifications that we can do. Thanks.

src/js/dtv-common-header.js
alex-deaconu: @rudibravo `svc-store-authorization.js` was added to `core-api-client`. Now **core**-api-client is debatably not the best place for this store module, but I can't think of a better spot for it.

So because it belongs to `core-api-client` I was just suggesting it follows the same module naming as other modules in that component.

Now regarding location, I would suggest maybe line 8 or line 16? Probably not a big deal either way as long as it's in that group.

Thanks.

rudibravo: @alex-deaconu now I understand what you mean. Sure I will fix that.
Thanks.

fjvallarino:
@rudibravo looks good. I'll review again after feature/display-licensing changes have been merged. Can I suggest merging instead of rebasing? It would make reviewing easier. Thanks!

rudibravo:
@alex-deaconu @fjvallarino addressed your comments and merged display-licensing back.
It's deployed to https://apps-stage-3.risevision.com

I noticed the plan bar is not updating. I debugged it and it's related to the Plan now being retrieved from the `company.get`. The Store is not calling the Webhook when starting a trial.

The plan modal still works fine because it retrieves the trial status from the store directly.

Thanks.

src/js/components/plans/svc-plan.js
```diff
@@ -101,6 +102,9 @@
               }
             }
           });
+          modalInstance.result.then(function () {
+            userState.refreshProfile();
```
alex-deaconu: @rudibravo as discussed, we're not sure if the Store webhook will be called in time to update the Company Settings with the Plan Trial info. I would suggest calling `userState.updateCompanySettings()` with an updated Company object that contains the Plan Trial information. @fjvallarino FYI. Thoughts? Thanks.

src/js/components/plans/svc-plan.js
```diff
@@ -130,6 +134,25 @@
...
+
+              deferred.resolve(plansMap);
+            })
+            .catch(function (err) {
+              deferred.reject(err);
+            });
+
+          return deferred.promise;
+        };
+
```
alex-deaconu: Note: this is undoing the changes made in the last PR.

src/js/components/plans/ctr-plans-modal.js
```diff
@@ -2,18 +2,26 @@ angular.module("risevision.common.components.plans")
...
     $scope.currentPlan = currentPlan;
+    $scope.startTrialError = null;
 
     $scope.getPlansDetails = function () {
       $loading.start("plans-modal");
 
-      return planFactory.getPlansDetails()
+      return planFactory.getCompanyPlanStatus()
+        .then(function (allPlansMap) {
+          $scope.allPlansMap = allPlansMap;
```
alex-deaconu: @rudibravo does this need to be a $scope variable? If not used in the partial, I'd rather just you use a local. Thanks.

alex-deaconu:
@rudibravo looks good, couple of comments. Thanks.

src/js/components/plans/svc-plan.js
rudibravo: @alex-deaconu good point, I actually overlooked the webhook constrain. I assume I can just update the plan with the company with the trial it just started, since it will have similar behaviour on the CompanySubscriptions.

alex-deaconu: @rudibravo yes, the duplicated logic is not ideal but I can't think of a better solution. Thanks.

src/js/components/plans/svc-plan.js
rudibravo: @alex-deaconu yes I discussed it with @fjvallarino and we decided to have a simple version of getCompany that still call the stores. The reason this PR needs this is because we need the subscription status for each plan from the store. Only listing the products is not enough to decide if a product can start trial or not.

alex-deaconu: @rudibravo understood 👍

src/js/components/plans/ctr-plans-modal.js
rudibravo: @alex-deaconu 👍

rudibravo:
@alex-deaconu addressed your comments. Thanks.

rudibravo:
@alex-deaconu found an issue so hold on to review this.

src/js/components/plans/ctr-plans-modal.js
```diff
@@ -8,13 +8,14 @@ angular.module("risevision.common.components.plans")
 
     $scope.currentPlan = currentPlan;
     $scope.startTrialError = null;
+    var _allPlansMap;
```
alex-deaconu: @rudibravo do you think we should maybe initialize to {} to prevent npes? Thanks.

alex-deaconu:
@rudibravo looks good, one minor suggestion 👍 . Thanks.

alex-deaconu:
@rudibravo 👍

PR #774, opened 2018-02-21
Use Store's DB status name 'Trial' instead of 'On Trial' from Subscription Service

fjvallarino:


rudibravo:
@fjvallarino 👍

alex-deaconu:
@fjvallarino @rudibravo would there be any reason to support both versions of the string? Thanks.

fjvallarino:
@alex-deaconu this only affects the CompanySubscriptions object. "Trial" is the value as stored in Store's DB, "On Trial" is how Subscription Status service returns it. We discussed this today on HipChat (Apps room, around 2pm your time). I'm fine either way.

alex-deaconu:
@fjvallarino okay, I saw the conversation, but was just curious. 👍 Thanks.

PR #776, opened 2018-02-23
Update Plans Modal to use planSubscriptionStatus field

fjvallarino:
When switching to using CompanySubscriptions entity instead of Subscription Service, a couple of fields stopped being available. Some statuses names have also changed (most notably, "Subscribed" changed to "Active" and "On Trial" to "Trial"). This also addresses the situation when Downgrade and Start Trial could be displayed at the same time.

We can discuss this tomorrow during Stand Up.


src/js/components/plans/ctr-plans-modal.js
```diff
@@ -38,7 +38,9 @@ angular.module("risevision.common.components.plans")
     };
 
     $scope.canUpgrade = function (plan) {
-      if (currentPlan.type === plan.type) {
+      if ($scope.canStartTrial(plan)) {
```
rudibravo: @fjvallarino since you added this check here, I think it's safe to remove from the [html](https://github.com/Rise-Vision/common-header/blob/159a37b052dcacffdc06d975075ef576b6fc2f33/src/templates/components/plans/plans-modal.html#L26)

fjvallarino: @rudibravo actually, I added it in the code because I didn't realize it was in the html. I'll remove it as you suggest.

fjvallarino:
@rudibravo updated!

rudibravo:
@fjvallarino looks good 👍

PR #779, opened 2018-02-28
Remove individual product trials

rudibravo:
@Rise-Vision/apps @fjvallarino update localization text for starting individual trials. also a bug where the downgrade button was showing incorrectly.

please review


fjvallarino:
@rudibravo looks good

PR #781, opened 2018-03-02
Fix issues with out of sync available pro licenses

fjvallarino:


src/js/components/userstate/services/svc-companystate.js
```diff
@@ -76,9 +76,31 @@
           return deferred.promise;
         };
 
+        var _reloadSelectedCompany = function () {
+          var deferred = $q.defer();
+
+          getCompany(_state.selectedCompany.id)
```
rudibravo: @fjvallarino I'm sure you had a reason to reload the selected company here, but I'm trying to understand why, since the trial status might still not be available before the webhook call is processed.

src/js/components/userstate/services/svc-companystate.js
```diff
@@ -76,9 +76,31 @@
...
+            .then(function (company) {
+              objectHelper.clearAndCopy(company, _state.selectedCompany);
+
+              deferred.resolve();
+              $rootScope.$broadcast("risevision.company.updated", {
+                "companyId": company.id
+              });
+            })
+            .then(null, function (resp) {
```
rudibravo: @fjvallarino just curious, any reason not to give the error callback on the first `then` or use the `catch` here?

src/js/components/plans/svc-plan.js
```diff
@@ -167,6 +167,33 @@
           return deferred.promise;
         };
 
+        _factory.getProLicenseCount = function (_company) {
+          var company = _company || userState.getCopyOfSelectedCompany(true);
```
rudibravo: @fjvallarino why are you using the `noFollow` here? I don't think it's required.

src/js/components/plans/svc-plan.js
```diff
@@ -167,6 +167,33 @@
...
+          return (company.planPlayerProLicenseCount || 0) + (company.playerProLicenseCount || 0);
+        };
+
+        _factory.areAllProLicensesUsed = function (_company) {
+          var company = _company || userState.getCopyOfSelectedCompany(true);
```
rudibravo: @fjvallarino again the `noFollow` parameter.

src/js/components/userstate/services/svc-companystate.js
fjvallarino: @rudibravo this is something we're testing with @ezequielc, mainly because we don't know which displays will be assigned a license if you already had some created and only then started a Trial (which means, we consider you have not used any license even though you may have been automatically assigned all of them). The idea is waiting for 5 seconds (this is done in the startTrial method in PlansModal), and only then reloading the company. It has worked fine with the company information itself, but we have not been able to validate the license assignment part, since it is not running on rvacore-test. 

This is clearly an awful workaround.

src/js/components/userstate/services/svc-companystate.js
fjvallarino: @rudibravo I just copied and modified the _switchCompany method, and I preferred to keep it as similar as possible. I guess the original implementer thought it was clearer to do it in two parts (similar to a separate catch clause)?

fjvallarino:
@rudibravo removed the nofollow parameter, and answered the comments

PR #791, opened 2018-03-29
Show Web Widget when clicking on Support button (restore old mechanism)

fjvallarino:
This is ready for review on https://apps-stage-1.risevision.com


alex-deaconu:
@fjvallarino I believe the Web Widget should show up for everyone, regardless of their Priority Support subscription. Current logic for `getSupport()` checks subscription and conditionally shows the Widget. 

@ezequielc can you please confirm what the expected behaviour is? Thanks!

ezequielc:
@alex-deaconu For now we just want to re-enable as it was. I think non-Priority Support users can use 'Send Us a Note'. 
We will revisit this flow soon. Thanks

alex-deaconu:
@ezequielc sounds good, thanks for the clarification. @fjvallarino looks good 👍 . Thanks!

PR #793, opened 2018-04-06
Fix replace issue in Email Template

alex-deaconu:
@ezequielc @olegrise please review. Thanks!


olegrise:
@alex-deaconu could you explain what was the problem? Why did you need to make this change?

alex-deaconu:
@olegrise more info on the issue here: https://trello.com/c/qljHZcYz/3391-issue-newuserencodedcompanyname-as-company-name-1. 

Basically, https://github.com/Rise-Vision/common-header/blob/master/src/templates/add-user-email.html has two instances of the "Sign Up" link. The string.replace function only replaces the first instance found in that String. Using the \*\g operator replaces all instances (https://www.w3schools.com/jsref/jsref_replace.asp).

olegrise:
@alex-deaconu makes sense. Looks good 👍

PR #810, opened 2018-05-14
Remove Need Support button

fjvallarino:
Requesting review of this PR, although we still have an issue (which exists in production too).

Conflict points of the PR are:

- Routes which require moving the Zendesk Help button up are checked explicitly in Common Header. Ideally, users should provide routes that need this behavior ('apps.editor.workspace' is hardcoded). I'm not sure it's worth the effort, that's why I did not pursue it (it's a simple change, anyway).
- Since the Zendesk script is now loaded on initialization, it causes issues with unit tests. The current fix consists on providing the script as an angular value (ZENDESK_WEB_WIDGET_SCRIPT), which is overriden for unit tests using a file added in gulpfile.js after all the other files are loaded. The main problem with the approach is that it requires the same logic in Apps (or Store, whenever we update the script).


src/templates/help-dropdown-menu.html
```diff
@@ -1,8 +0,0 @@
-
-<ul>
-    <li><a id="getStartedButton" href="https://help.risevision.com/hc/en-us/articles/115002868706-Get-started-with-Rise-Vision" target="_blank">Get Started</a></li>
```
alex-deaconu: @fjvallarino to confirm, we're removing this menu all together? Thanks.

alex-deaconu:
@fjvallarino don't forget bump versions before merging, also I'd be ok with going to **v2.4.0** for this release. Thanks.

src/templates/help-dropdown-menu.html
fjvallarino: @alex-deaconu that's my understanding based on this document: https://docs.google.com/document/d/1QyylMyAbxuYY9bURCpovd-vOC82m0Es3_dS3snVo42I/edit

@RV-Adrian FYI

alex-deaconu: @fjvallarino sounds good, just double checking. Thanks!

test/unit/services/svc-zendesk-override.js
```diff
@@ -0,0 +1,9 @@
+/* jshint maxlen: false */
+
+(function (angular) {
+  "use strict";
+
+  angular.module("risevision.common.support")
+  /* jshint quotmark: single */
+  .value('ZENDESK_WEB_WIDGET_SCRIPT', '');
```
alex-deaconu: @fjvallarino curious, why does this require single quotes?

fjvallarino: @alex-deaconu it doesn't, I didn't even think about it when I copied and replaced the inner text. I'll remove that and use double quotes.

src/js/controllers/ctr-help-buttons.js
```diff
@@ -1,31 +0,0 @@
-angular.module("risevision.common.header")
-
-.controller("HelpDropdownButtonCtrl", ["zendesk", "$scope", "supportFactory",
```
alex-deaconu: @fjvallarino is `supportFactory` still needed/used? Thanks.

test/unit/services/svc-zendesk-override.js
alex-deaconu: @fjvallarino 👍

src/js/services/svc-zendesk.js
```diff
@@ -165,13 +147,10 @@
               var rvCompanyInput = iframe.contents().find(
                 "input[name=24893323]");
               if (rvCompanyInput && rvCompanyInput.length > 0) {
-                getSubscriptionStatus().then(function (
-                  subscriptionStatus) {
+                getSubscriptionStatus().then(function (subscriptionStatus) {
```
alex-deaconu: @fjvallarino do you think we can re-use a generic Subscription Status service for this, instead of this custom, badly named `getSubscriptionStatus` service? Thanks.

src/js/controllers/ctr-help-buttons.js
fjvallarino: @alex-deaconu good point. It's required by /support and /send-note routes in Apps (which do not really work anymore, since the corresponding modals have been removed). Not sure what to do about them (i.e., if they should be removed or the should take a different action). 

@RV-Adrian do you have something in mind for those routes?

src/js/services/svc-zendesk.js
fjvallarino: @alex-deaconu actually, the only part we need of that function/service is 

```subscriptionStatusService.get(SUPPORT_PRODUCT_CODE, userState.getSelectedCompanyId())```

The rest is just a wrapper over that call, and subscriptionStatusService is the service we use everywhere else.

fjvallarino: @alex-deaconu maybe we can rename it to _getSupportSubscriptionStatus_?

alex-deaconu: @fjvallarino makes sense 👍

src/js/controllers/ctr-help-buttons.js
alex-deaconu: @ezequielc FYI, maybe you have some feedback for the above question as well. Thanks.

src/templates/help-dropdown-menu.html
RV-Adrian: @fjvallarino 👍

src/js/controllers/ctr-help-buttons.js
alex-deaconu: @RV-Adrian @fjvallarino from what I remember, those URLs were needed to forward users from the Website. However, since the website has its own Zendesk widget, are those links still there and are they still needed? Thanks.

ezequielc: @RV-Adrian @fjvallarino I agree with @alex-deaconu, these URLs are no longer needed since we removed the links from the website. @fjvallarino You can remove them. Thanks!

fjvallarino:
@alex-deaconu updated

PR #814, opened 2018-05-17
Check Pro Subscription status when calculating available licenses

fjvallarino:
Related to: https://trello.com/c/rlh24Dw3/3597-issue-382-cancelled-player-pro-subscription-still-counts-as-active-rpp-status-not-taken-into-account-1

Not bumping version yet since I think a previous PR is going to be merged before.


fjvallarino:
@alex-deaconu I just rebuilt to have the tests pass. Please review

src/js/components/plans/svc-plan.js
```diff
@@ -163,15 +163,17 @@
 
         _factory.getProLicenseCount = function (_company) {
           var company = _company || userState.getCopyOfSelectedCompany();
-          return (company.planPlayerProLicenseCount || 0) + (company.playerProLicenseCount || 0);
+          var proSubscribed = company.playerProSubscriptionStatus === "Active";
```
alex-deaconu: @fjvallarino we have a `_factory.isProSubscribed` function in this service; should we be re-using that? I see this takes a `_company` parameter here so maybe the other function is not useful... Thanks.

src/js/components/plans/svc-plan.js
```diff
@@ -163,15 +163,17 @@
...
+
+          return (company.planPlayerProLicenseCount || 0) + ((proSubscribed && company.playerProLicenseCount) || 0);
```
alex-deaconu: @fjvallarino does `company.planPlayerProLicenseCount` calculation take into account the Company Plan Subscription Status? Thanks.

alex-deaconu:
@fjvallarino couple of clarifications inline. Otherwise we should be good to go. Thanks.

src/js/components/plans/svc-plan.js
fjvallarino: @alex-deaconu yes, the _company parameter was the reason I did not reuse the other one. I thought about refactoring a bit, mainly to avoid duplicating the validation, but I was not sure it was worth it.

alex-deaconu: @fjvallarino sounds good. Thanks.

src/js/components/plans/svc-plan.js
fjvallarino: @alex-deaconu mmm, maybe it should, yes. My reasoning was that the current plan is necessarily active, otherwise you'd be moved to a lower plan, but I think for Suspended status that does not really happen (you are still on that plan for a while)

fjvallarino: @alex-deaconu actually, I mentioned that in the Trello card and during Stand Up, not in the PRs header.

alex-deaconu: @fjvallarino okay, I'm not really aware of all the use cases for this. Where does the lower plan move happen? Is it in the Core, as I don't see it anywhere in this service. Thanks.

fjvallarino: @alex-deaconu yes, that's on Core's side. If your plan is Suspended, the endpoint will still return that Plan (it's different if your plan is expired or trial-expired, in which case it returns a lower plan). There's no explicit handling of Suspended in svc-plan.js, since it checks for the valid cases instead (Active, On Trial). The only place where we explicitely handle that status is here:

https://github.com/Rise-Vision/common-header/blob/master/src/templates/plan-banner.html

alex-deaconu: @fjvallarino okay. It's very confusing; I guess it works as intended so we're good. Thanks.

ezequielc:
@fjvallarino 👍

PR #815, opened 2018-05-24
Add help button in header for logged in users

fjvallarino:
I understood we're leaving the Priority Support removal for a future card, so I'm requesting review as is.


src/js/controllers/ctr-zendesk-button.js
```diff
@@ -0,0 +1,9 @@
+angular.module("risevision.common.header")
+
+.controller("ZendeskButtonCtrl", ["$scope", "userState", "zendesk",
```
alex-deaconu: @fjvallarino I don't think userState is used here. Thanks.

src/scss/core/global.scss
```diff
@@ -305,9 +305,9 @@ ul.social {
 .btn-default svg {
   margin-left: 5px;
 }
-.btn svg path {
-  fill: #666;
-}
+// .btn svg path {
+//   fill: #666;
+// }
```
alex-deaconu: @fjvallarino @RV-Adrian should we remove this code rather than comment it out? Thanks.

alex-deaconu:
@fjvallarino 2 minor cleanup comments. Otherwise PR looks good. Thanks for the fixes 👍

src/scss/core/global.scss
RV-Adrian: @fjvallarino @alex-deaconu 
Yes, let's remove it so we can deploy. I couldn't find a purpose for it (possibly for an old button style)

alex-deaconu:
@fjvallarino looks good 👍

PR #816, opened 2018-05-24
Add two seconds delay before showing Zendesk button, to avoid race conditions on some page reload situations

fjvallarino:


src/js/components/loading/svc-loading.js
```diff
@@ -26,9 +26,11 @@ angular.module("risevision.common.components.loading")
...
       /* Global Spinner */
       //append global spinner
-      angular.element($document[0].body).append(
-        "<div rv-global-spinner class=\"ng-hide\" style=\"position: fixed; width: 100%; height: 100%; top: 0; left: 0; z-index: 1040; \"></div>"
-      );
+      if ($document && $document[0]) {
+        angular.element($document[0].body).append(
+          "<div rv-global-spinner class=\"ng-hide\" style=\"position: fixed; width: 100%; height: 100%; top: 0; left: 0; z-index: 1040; \"></div>"
+        );
+      }
```
alex-deaconu: @fjvallarino are these and the fastpass changes required to fix unit tests as well? Thanks.

src/js/services/svc-zendesk.js
```diff
@@ -227,9 +233,11 @@
         }
 
         function _showWebWidget() {
-          zendesk.logout();
-          zendesk.displayButton();
-          widgetVisible = true;
+          setTimeout(function () {
```
alex-deaconu: @fjvallarino curious why not use the $timeout service instead? Thanks.

alex-deaconu:
@fjvallarino looks good 👍 . Thanks!

src/js/components/loading/svc-loading.js
fjvallarino: @alex-deaconu the changes in svc-loading.js were needed to fix the issues caused by adding the dependency to userAuthService on the ```run``` function of svc-zendesk.js (two tests were failing on svc-primus.tests.js). Since it was pretty hard to track down I also decided to add those checks in svc-fastpass.js, although they are not really needed for this fix.

src/js/services/svc-zendesk.js
fjvallarino: @alex-deaconu no particular reason

src/js/components/loading/svc-loading.js
alex-deaconu: @fjvallarino makes sense, thanks!

alex-deaconu:
@fjvallarino I'm going to merge this PR so we can release this fix today before noon. Thanks.

PR #819, opened 2018-05-30
Additional Error Logging

alex-deaconu:
Log error if Auth fails after redirect
Log error for automatic auth change detection

@fjvallarino follow up PR with additional logging. No functionality changes.

Please review. Thanks!


fjvallarino:
@alex-deaconu looks good 👍

PR #821, opened 2018-06-06
Check connection if client api doesn't load

alex-deaconu:
Check server connection URL
if the Client API doesn't load within 10 seconds

Throw an error if the Server doesn't respond
with a 404 (expected response) or other valid responses

Update Storage to use v2 API
Updated i18n strings & cleanup


alex-deaconu:
Adding a verification check for gapi.client.load functionality. If the API doesn't load within 10 seconds (and it has a URL, ie. Core/Store/Storage) this code will double check if the server at that particular URL is accessible. The expected response is 404, but valid 2xx or 3xx responses are allowed as well.

If an invalid response is received (encountered statuses have been -1 and 403) we throw a custom error. It is the onus of the application to digest and show the error properly.

Also included are updated i18n strings for Storage errors and Storage i18n refactor. See Apps PR for that.

src/locales/locales_po/en/storage-client.po
```diff
@@ -10,19 +10,19 @@ msgstr "If you are using this asset in any Presentation, the link for each must
...
-msgid "delete-files-plural"
+msgid "delete.files-plural"
 msgstr "{{count}} files are about to be permanently deleted."
 
-msgid "delete-files-singular"
+msgid "delete.files-singular"
 msgstr "One file is about to be permanently deleted."
 
-msgid "delete-folders-plural"
+msgid "delete.folders-plural"
```
fjvallarino: @alex-deaconu are these translation strings updated in some rise-vision-apps PR? I don't see it in https://github.com/Rise-Vision/rise-vision-apps/pull/660

PR #830, opened 2018-06-19
scrollbar-fix update

ghislaineguerin:
STAGE LINK: https://apps-stage-4.risevision.com
Please review.


src/scss/ui-components/common-header.scss
```diff
@@ -9,6 +9,9 @@
   z-index: 1000;
   margin-bottom: 0;
   &.navbar-default {
+    .navbar-nav>.open>a {
```
RV-Adrian: @ghislaineguerin I noticed a flash of blue here on my local build when the dropdown opens. This fixes it, right?

ghislaineguerin: @RV-Adrian yes! it was a default bootstrap state so I had to add the overwrite again.

RV-Adrian: @ghislaineguerin ok perfect!

PR #833, opened 2018-07-06
add worksheet warning and update data error message

santiagonoguez:


alex-deaconu:
@santiagonoguez you are missing the build results in the `dist` folder, you can run `gulp i18n-build` (or `gulp build`) to obtain that. Thanks.

alex-deaconu:
@santiagonoguez looks good, thanks for fixing 👍

PR #834, opened 2018-07-09
fix invalid range message

santiagonoguez:


alex-deaconu:
@santiagonoguez you're only updating the tag in package.json? what about bower.json? Thanks.

santiagonoguez:
@alex-deaconu corrected, thanks !

alex-deaconu:
@santiagonoguez thanks for fixing 👍

PR #841, opened 2018-07-19
Billing Address validation

alex-deaconu:
Only validate current page


alex-deaconu:
Adding validation for the Billing address fields. Some things to note:
- Unit field is not required
- Navigation (Continue and Back buttons) is blocked until the form is valid
- Validation is triggered when a field is edited
- Because of this, the validation message at the top of the page will not appear so I didn't add it

Because all fields just have required field validation, you need to type something and delete it for the field to show as invalid. Please review the stage and let me know if you have any suggestions for improving the validation process or this works.

Staged on https://apps-stage-3.risevision.com

olegrise:
@alex-deaconu why "Back" button is [greyed out](https://github.com/Rise-Vision/common-header/pull/841/commits/6334e125a209d11615b70466e508364c356f3c42#diff-1bee1e3eb134cd521f429fecf72e4a35R93) ? :)

alex-deaconu:
@olegrise not sure, mostly for consistency. I can remove. Anyone else have an opinion on this? Thanks.

olegrise:
@alex-deaconu the way I see it that you should be able to go back regardless if your address is valid or not.

ezequielc:
@alex-deaconu Maybe the staging failed, but the Back button is still disabled for me. 
Also, on the auto complete, the city is not auto completing for me. Thanks!

olegrise:
@alex-deaconu re. back button - thanks, looks good. What is the plan for validating dropdown fields?

alex-deaconu:
@Rise-Vision/apps I've re-worked the PR to allow the "Continue" button to be clicked. This will automatically trigger validation.

To confirm, validation will be triggered when a field is modified, or when the Continue button is clicked.

Unfortunately Circle CI is still down so I cannot stage.

@ezequielc the staging environment is not updated with the latest autocomplete properties, as per the above.

Please review functionality, will post a staging links once I have it. Thanks.

ezequielc:
@alex-deaconu Sounds good. Thanks

alex-deaconu:
The latest version of this PR is staged at https://apps-stage-3.risevision.com. To confirm, it has changes for the following:
- Fixes for the autocomplete field attributes
- Enabling the back button even if the form is invalid
- Enabling Continue and Showing validation errors when Continue is clicked

Please review. Thanks!

olegrise:
@alex-deaconu when I click Continue, the form allows me to proceed without selecting province. I had Canada selected as country. I think the rules should be that if Canada or USA are selected, then province/zip is required. For any other country it should not be required.

alex-deaconu:
@olegrise good catch. I've updated the branch to clear the value when switching between countries (`ng-if`/`ng-init` implementation).

Also I removed validation from the Province when the Country is not CA/US, and the telephone number as discussed.

Finally, I fixed the back button for the next page (shipping address) to allow easier review (and cleaned up the unnecessary fields).

Re-staged at stage-3. Please review. Thanks!

ezequielc:
@alex-deaconu Looks good, the only thing I notice is that Province/State is not highlighted in red when required. It shows the error on the top of the modal, but doesn't set the field to red. Thanks

alex-deaconu:
@ezequielc yes, there are still some issues with the validation that I'm trying to work through. My previous commit introduced the "not highlighted" state you're referring to. Will update once fixed. Thanks.

alex-deaconu:
@ezequielc @olegrise my latest commit should address all the issues with the address validation.

To summarize:
Address will not validate if Country is blank or other than CA/US.
For either CA or US, it will validate blank values, and invalid province/state values. To replicate, type a province in the text box and then select CA/US, or select a value from CA Province dropdown and then switch to US (or vice versa).

Also, included validation for field lengths as per Avalara v2.0 API requirements (https://github.com/Rise-Vision/store/blob/master/web/script/services/svc-address.js#L134)

ezequielc:
@alex-deaconu Looks good to me. Thanks

olegrise:
@alexey-rise good stuff. The only thing that is still not working is autocomplete for Province, but I think we can leave with this. Well, assuming support will not think it's a bug 😄.

alex-deaconu:
@olegrise not sure why it's not working for you. It worked for me with 2 different addresses I have saved in Chrome.

olegrise:
@alex-deaconu is it working for you if country = Canada?

src/templates/components/purchase-flow/checkout-billing-address.html
```diff
@@ -1,75 +1,76 @@
...
       </div>
     </div>
     <div class="row">
       <div class="col-md-6">
-        <div class="form-group">
+        <div class="form-group" ng-class="{ 'has-error': isFieldInvalid('province') }">
           <label for="billing-address-region" class="control-label">State/Province/Region</label>
 
-          <input type="text" class="form-control" ng-model="plan.billingAddress.province" autcomplete="province" ng-hide="plan.billingAddress.country === 'US' || plan.billingAddress.country === 'CA'" />
+          <input name="province" type="text" class="form-control" ng-model="plan.billingAddress.province" autcomplete="address-level1" ng-show="plan.billingAddress.country !== 'US' && plan.billingAddress.country !== 'CA'" province-validator="plan.billingAddress.country">
```
alex-deaconu: @olegrise this would probably explain it: `autcomplete=address-level1` (note the typo). Not sure why they still worked for me. Will fix, sorry about that. Thanks for sticking with it!!

PR #842, opened 2018-07-19
Fix yearly price calculation for additional displays

alex-deaconu:
Unfortunately CircleCI is down and this didn't build. Will retry once the outage is over.


fjvallarino:
@alex-deaconu looks good 👍

PR #843, opened 2018-07-19
Update customer portal links to account for Chargebee companies

fjvallarino:


fjvallarino:
@alex-deaconu added a few commits (starting on 23c6cb8) which add Chargebee integration in the same tab, remove no longer used service, and adds the missing unit tests.

src/js/services/svc-chargebee-factory.js
```diff
@@ -0,0 +1,136 @@
...
+angular.module("risevision.store.services")
+  .factory("getChargebeeInstance", ["$q", "storeService",
+    function ($q, storeService) {
+      var currentCompanyId = null;
+      var currentInstance = null;
+
+      function _createChargebeeInstance(session) {
+        var cbInstance = {};
+
+        cbInstance.instance = Chargebee.init({
```
alex-deaconu: @fjvallarino I think you should be using $window.Chargebee instead of the global variable. Thanks.

src/js/services/svc-store-service.js
```diff
@@ -0,0 +1,54 @@
+(function () {
+  "use strict";
+
+  angular.module("risevision.store.services")
+    .service("storeService", ["$q", "$log", "storeAPILoader",
+      function ($q, $log, storeAPILoader) {
+        var service = {
+          openPortal: function (companyId, returnUrl) {
```
alex-deaconu: @fjvallarino to confirm, this is not currently used as we use the Portal API? Thanks.

src/js/services/svc-chargebee-factory.js
```diff
@@ -0,0 +1,136 @@
...
+        if (currentCompanyId === companyId) {
+          return $q.resolve(currentInstance);
+        } else {
+          var deferred = $q.defer();
+
+          storeService.createSession(companyId)
+            .then(function (session) {
+              console.log("Chargebee session for companyId", companyId, "is", session);
+
+              currentInstance = _createChargebeeInstance(session);
```
alex-deaconu: @fjvallarino what happens when this session expires? Do we need to grab another token? How do we know to do that? Thanks.

src/templates/components/plans/plans-modal.html
```diff
@@ -63,9 +67,12 @@ <h3>{{plan.proLicenseCount}}</h3>
             <a id="subscribe-plan" ng-show="getVisibleAction(plan) === 'subscribe'"
                 target="_blank" href="https://store.risevision.com/product/{{plan.productId}}"
                 class="btn btn-primary btn-block" translate>common-header.plans.subscribe</a>
```
alex-deaconu: @fjvallarino looking at this, it seems I was mistaken in suggesting that we have an "Upgrade" button. Both scenarios (for Subscribe and Upgrade) lead to this path.

However, I think we need to split this path into 2 separate use cases. Purchasing a Plan (from Trial or inactive Subscription) should show the In App Payment, and Upgrading to a new Plan while on an active Plan, which should lead to the Portal. Thoughts? 

Thanks!

src/js/services/svc-store-service.js
fjvallarino: @alex-deaconu I think removing makes sense, yes. I left it there because some cards mention Chargebee URL, but I guess in the end we will prefer opening In App

src/js/services/svc-chargebee-factory.js
fjvallarino: @alex-deaconu sure, no problem. Is there any advantage?

src/js/services/svc-chargebee-factory.js
fjvallarino: @alex-deaconu the session is supposed to be good for one hour. Do you think launching a timer and invalidating the session object (in order for it to be created again the next time we request something Chargebee related) is good enough? Sounds like a simple solution (we don't have a lot of control when it comes to error handling of Chargebee functions)

src/templates/components/plans/plans-modal.html
fjvallarino: @alex-deaconu makes sense (both your suggestion about each path, and handling the changes in a separate PR)

src/js/services/svc-chargebee-factory.js
alex-deaconu: @fjvallarino not really, just consistency with other places where we use $window objects. Thanks for updating.

src/js/services/svc-chargebee-factory.js
alex-deaconu: @fjvallarino is there any way to get that information from the library? ie. get a callback or call a function that would tell us when the session is expired? Thanks.

src/js/services/svc-chargebee-factory.js
```diff
@@ -0,0 +1,136 @@
...
+  .factory("getChargebeeInstance", ["$q", "storeService",
+    function ($q, storeService) {
+      var currentCompanyId = null;
+      var currentInstance = null;
+
+      function _createChargebeeInstance(session) {
+        var cbInstance = {};
+
+        cbInstance.instance = Chargebee.init({
+          site: "risevision-test"
```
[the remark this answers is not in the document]
fjvallarino: @olegrise @alex-deaconu for some reason I can't comment on the suggestion below:

*I was thinking the same actually. Ideally we should create session every time customer clicks on a link. There are 2 reasons: 1) session can expire 2) customer may not exist in CBee when they opened the App, then they may purchase a plan and click on the Account page.*

I don't see a way for adding error handling on the ```openPortal``` and ```openSection``` methods (https://www.chargebee.com/checkout-portal-docs/api.html#chargebee-portal-instance-object), which hinders our chances of renewing the session on error. 

I added the required logic to take into account the expiration time being sent as part of Chargebee's session object, which should be enough to avoiding issues of session expiration (item 1). Regarding item 2, the call to get a session object from the Store will fail if the customer does not exist on Chargebee. This is something I will also be handling in a follow up card (we should send them to Store until a Chargebee account is created through Subscription or Trial)

.jshintrc
```diff
@@ -77,6 +77,7 @@
     "Stripe": false,
     "Spinner": false,
     "SignInBaseCtr": false,
+    "Chargebee": false,
```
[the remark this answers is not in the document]
fjvallarino: @alex-deaconu updated

src/js/services/svc-chargebee-factory.js
fjvallarino: @alex-deaconu @olegrise the new version uses the information that comes from Chargebee's session object

src/js/services/svc-chargebee-factory.js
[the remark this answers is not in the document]
fjvallarino: @olegrise @alex-deaconu by the way, the Chargebee instance is now selected based on the environment's config file

src/js/services/svc-chargebee-factory.js
```diff
@@ -0,0 +1,145 @@
...
+      function _isSessionExpired() {
+        // Leaves a 1 minute buffer to avoid expiration on call
+        return currentSessionExpiration - Date.now() < 60000;
+      }
+
+      function _createChargebeeInstance(session) {
+        var cbInstance = {};
+
+        cbInstance.instance = $window.Chargebee.init({
+          site: CHARGEBEE_INSTANCE
```
olegrise: @fjvallarino will this work for the test company on production?
@alex-deaconu

fjvallarino: @olegrise I'm not sure I understand. That constant comes from the config file of each environment. For both ```beta``` and ```prod```, that value is defined as ```risevision```. For the other environments (dev and stage) it is ```risevision-test```.

olegrise: @fjvallarino my original comment was "the site for the **test company** is "risevision-test" and "risevision" is for production company." 

In the test environment (`rvacote-test`server) all companies are test companies. Any call from `rvacote-test` should go to `risevision-test` in Chargebee. 

In the production environment (`rvaserver2` server) some companies are production companies and some are test companies. If company is the TEST company, then app should use `risevision-test`site. If company is the PRODUCTION company, then app should use `risevision`site. Do you follow me?
@alex-deaconu

fjvallarino: @olegrise I get it now. Do we have any way of knowing if a company is test or production?

fjvallarino: @olegrise updated. It chooses the environment based on ```userState.isTestCompanySelected```, but for Stage both constants point to ```risevision-test```

olegrise: @fjvallarino Thanks! 👍

olegrise:
@fjvallarino I need more time to finish my review, so please don't merge it yet. One question that I want to have a discussion about is if we should be using ID of the user company or selected company when opening customer portal. I'm not confident that selected company is the correct answer. Let's talk about it tomorrow.
@alex-deaconu @ezequielc

src/js/config/dev.js
```diff
@@ -39,5 +39,7 @@
     .value("SUPPORT_PRODUCT_ID", "14")
     .value("SUPPORT_PRODUCT_URL",
       "https://store.risevision.com/products/?cat=compareSupport")
-    .value("APPS_URL", "https://apps-stage-0.risevision.com");
+    .value("APPS_URL", "https://apps-stage-0.risevision.com")
+    .value("CHARGEBEE_TEST_INSTANCE", "risevision-test")
```
olegrise: @fjvallarino the more I look at this, the more I find the `INSTANCE` suffix confusing. In your code you have other functions use word "instance" and they reference something completely different. My recommendation is to use suffix `SITE` i.e.  "CHARGEBEE_TEST_SITE" and "CHARGEBEE_PROD_SITE". At least this way it will match how Chargebee calls it [here](https://github.com/Rise-Vision/common-header/pull/843/files#diff-2322747728e2c7a5435e6361836e3504R20).

fjvallarino: @olegrise updated

src/js/services/svc-chargebee-factory.js
```diff
@@ -0,0 +1,146 @@
...
+  .factory("getChargebeeInstance", ["$q", "$window", "storeService", "userState",
+    "CHARGEBEE_TEST_SITE", "CHARGEBEE_PROD_SITE",
+    function ($q, $window, storeService, userState, CHARGEBEE_TEST_SITE, CHARGEBEE_PROD_SITE) {
+      var currentCompanyId = null;
+      var currentInstance = null;
+      var currentSessionExpiration = 0;
+
+      function _isSessionExpired() {
+        // Leaves a 1 minute buffer to avoid expiration on call
+        return currentSessionExpiration - Date.now() < 60000;
```
olegrise: @fjvallarino we discussed this in standup today. Curious, have you tested what happens if a user opens portal 1 minute before session expires? Will the user be kicked out of the portal in 1 minute?
@ezequielc @alex-deaconu

olegrise:
@fjvallarino re. "should we be using ID of the user company or selected company when opening customer portal" - we discussed this during standup. It was decided that we are going to leave the implementation as is for now i.e. use ID of the selected company. Thanks.
@alex-deaconu @ezequielc

src/js/services/svc-chargebee-factory.js
fjvallarino: @alex-deaconu @ezequielc @olegrise I tried opening Apps, clicking on "Account and Details" (so it would open Customer Portal), closing it, then re-opening it 57-58 minutes later, and it worked fine. Left Customer Portal open for about two hours and could still use it without issues.

PR #845, opened 2018-07-23
Populate billing, shipping and contact from userState

alex-deaconu:
Populate with the same fields as the Store checkout. Also, clean out contact object as per the Store code as well.


fjvallarino:
@alex-deaconu looks good 👍

olegrise:
👍 better late than never 😄

PR #848, opened 2018-07-27
Shipping Address form & Address validation step

alex-deaconu:
Split Address Fields in separate directive

Validate address with the Store server
Show validation errors if any


alex-deaconu:
PR includes changes for Address Validation with the Store server from the previous PR.

I've updated the error message for when a different address is returned as suggested. However, I would like to keep this as a `$log.error` for now just to bring attention to cases where the response is different from the original address. Technically we should not encounter this error.

src/templates/components/plans/plans-modal.html
```diff
@@ -56,8 +56,8 @@ <h3>{{plan.proLicenseCount}}</h3>
...
           <p class="small u_margin-sm-bottom text-danger" ng-show="isCurrentPlan(plan) && isTrialExpired(plan)" translate>common-header.plans.trial-expired</p>
 
-          <a id="subscribe-plan" ng-if="origin === 'Rise'" ng-show="getVisibleAction(plan) === 'subscribe'" class="btn btn-primary btn-block" target="_blank" href="https://store.risevision.com/product/{{plan.productId}}" translate>common-header.plans.subscribe</a>
-          <a id="subscribe-plan" ng-if="origin !== 'Rise'" ng-show="getVisibleAction(plan) === 'subscribe'" class="btn btn-primary btn-block" ng-click="purchaseFactory.showPurchaseModal(plan, isMonthly)" translate>common-header.plans.subscribe</a>
+          <a id="subscribe-plan" ng-if="origin === 'rise'" ng-show="getVisibleAction(plan) === 'subscribe'" class="btn btn-primary btn-block" target="_blank" href="https://store.risevision.com/product/{{plan.productId}}" translate>common-header.plans.subscribe</a>
+          <a id="subscribe-plan" ng-if="origin !== 'rise'" ng-show="getVisibleAction(plan) === 'subscribe'" class="btn btn-primary btn-block" ng-click="purchaseFactory.showPurchaseModal(plan, isMonthly)" translate>common-header.plans.subscribe</a>
```
alex-deaconu: @olegrise yes, this is still here. I'm just using this to basically ignore the origin field and always show the checkout form. To confirm, we need to update this to use the 
`userState. isSelectedCompanyChargebee()`
function that is in Francisco's CH PR so I would like to leave it in for ease of testing.

Thanks.

olegrise: @alex-deaconu 👍 😸

alex-deaconu:
@olegrise fixed issue we discussed. Checking if form is valid before calling server side address validation.

Still investigating a fix for the spinner issue.

Thanks.

olegrise:
@alex-deaconu re. fining loading spinner - see [here](https://github.com/Rise-Vision/store/blob/master/web/script/ctr-admin-company-settings-modal.js#L9-L17) and [here](https://github.com/Rise-Vision/store/blob/master/web/view/admin-company-settings-modal.html#L2). Basically you set the sinner to autostart in HTML and then you turn it off in your code right away.
P.S. - one day we should fix that spinner directive 😠

olegrise:
@alex-deaconu couple of findings:
- On the Shipping Address page - when I leave Company Name empty, it allows me to continue to the next page.
- On the Shipping Address page - when I leave Street Address empty, it seems that the page is doing the server side validation instead of just telling me about the problem right away.
- Autocomplete is still not working as good as is should. See [video](https://drive.google.com/file/d/1IoezS6d5Eq-xgr5AxBGvvufqAQv4bbMh/view?usp=sharing).

alex-deaconu:
@olegrise re: validation, I actually forgot to add the form to the validators. It should be fixed in the latest commit.

Re: auto-complete, the one thing I found is that in previous forms you had "Ontario" as the province (instead of ON), it will not find the value. Wondering if you could change the country to something else than CA/US and check the value in the Province Text Box. 

Thanks!

PR #855, opened 2018-08-07
Load Stripe js and create CC token

alex-deaconu:
Load stripe (requires script tag in html)
Set key conditionally on environment

Check card fields with Card API
Create token via Stripe API


alex-deaconu:
As described above, additional CC fields are being validated.

Card number, CVC and Expiry Date are validated with the Stripe.js library. Card number is validated with the Luhn check (https://en.wikipedia.org/wiki/Luhn_algorithm). CVC has to be a number with 3-4 digits. Date has to be in the future.

If this validation succeeds, we call the Create Token API. So far, I haven't been able to get that API to fail. Apparently it always succeeds if using a Stripe Test account which is what we have configured for Staging and Test Companies. The API returns a token, however nothing is done with that token at this point. It just forwards to the next page.

To note, in Oleg's prototype, isTest is defined from the User's Company (https://github.com/Rise-Vision/rise-vision-apps/pull/634/files#diff-c7a194abcffddcb055046901cf1088c0R139). My implementation uses the Selected Company's isTest value. I would just like to confirm this is correct.

src/js/components/purchase-flow/svc-stripe.js
```diff
@@ -0,0 +1,93 @@
...
+          if (!$window.Stripe.card.validateCardNumber(card.number)) {
+            errors.push(STRIPE_ERRORS.invalid_number);
+          }
+          if (!$window.Stripe.card.validateCVC(card.cvc)) {
+            errors.push(STRIPE_ERRORS.invalid_cvc);
+          }
+        }
+
+        if (!$window.Stripe.card.validateExpiry(card.expMonth, card.expYear)) {
+          errors.push("Invalid Expiry Date.");
```
fjvallarino: @alex-deaconu should this error be in STRIPE_ERRORS too?

alex-deaconu: @fjvallarino my impression is that the STRIPE_ERRORS list is a list of error strings that can be received from Stripe. And this error is not included in those strings (there are separate month/year expiry errors). In order to keep that distinction, I kept this String separate.

This assumption is based on @olegrise 's prototype code (https://github.com/Rise-Vision/rise-vision-apps/pull/634/files#diff-9d7a986e2e6c42a99a8642009193f97aR29).

@olegrise can you please confirm if this is the case?

Thanks!

fjvallarino:
@alex-deaconu code looks good 👍. Based on your comment, will we be adding a script tag in Apps' index.html similar to the one for Chargebee?

alex-deaconu:
@fjvallarino yup, the Apps changes are pretty straight forward: https://github.com/Rise-Vision/rise-vision-apps/compare/feature/purchase-flow?expand=1. Thanks.

PR #856, opened 2018-08-08
Review Purchase UI Implementation

alex-deaconu:
Removed validateAddress service
Moved to storeService

Created various filters

Implemented calculateTaxes functionality

----

Moved purchase business logic to factory
Moved from controller to factory
Re-named the main data variable

Use user company isTest value
Rename section to Purchasing For
Remove Email reference


alex-deaconu:
To confirm, everything on the Purchase Review step except the Pay Now button should be implemented.

Also, added logic that allows the user to Edit any of the previous forms, and return directly to that page (without paging through the rest). The Estimate/Tax Calculation will be refreshed in this case.

One issue I would like to note is that I did not find any suitable way to show Errors related to the Estimate/Tax calculation, or to trigger a Retry. I'm not sure if that was part of the UX implementation.

Staged at https://apps-stage-3.risevision.com.

olegrise:
@alex-deaconu I reviewed. Couple of comments:
- When I click "Continue" on the Shipping Address page, the spinner starts and then closes too soon. I mean it closes and then I can still see the Shipping Address page for a second or two. At one point I even clicked "Continue" second time wondering why it didn't go the next page.
- When I accidentally click outside of the modal, the modal closes. I know it's not related to this PR, but I think this is not desirable behaviour for this modal. We can discuss this point at standup.

alex-deaconu:
@olegrise if you're referring to going to the last page, then opening the shipping address, and clicking continue, then yes, I was able to replicate. I know what the issue is and I will fix tomorrow with the improvements for displaying the Error Messages.

Otherwise, if you're completing the step for the first time, it should not behave in that manner.

In terms of closing the Modal, that is an easy fix. Will address as well.

Thanks!

alex-deaconu:
@olegrise I've fixed the issues you have highlighted in your comment. To confirm, the modal should no longer close when clicking outside of it. Also, I fixed the spinner issue with the Address, in 2 places. First off now, the page is switched right away when reaching the Purchase Review. Second, I fixed the spinner overwrite which was happening after the page switch on Address validation.

Also in this PR is the merge from Ghislaine's styling changes. Together with that I've added asterisks for required fields.

Re-staged at https://apps-stage-3.risevision.com.

@Rise-Vision/apps FYI. Please review the last few commits. Thanks!

olegrise:
@alex-deaconu fixes looks good. Here is the new issue :) - I clicked on "Edit" Payment Method couple of times and then I've got this:
screen shot 2018-08-10 at 11 40 19 am

The problem is that I cannot edit m card.

alex-deaconu:
@olegrise hmm, that's kinda the idea. Once we generate the token we clear the Credit Card information and keep only the generated token.

What I can do is enable the dropdown for selecting a Credit Card. This way they can keep using that card or add a new one. Does that work?

@ezequielc FYI. Thanks.

alex-deaconu:
@olegrise I've updated the code to show the Credit Card selector modal. Once you return to the Credit Card form, you can select to keep using that Card or add a new one. I think this should address your concern re: editing.

Also added a Retry button as discussed for Tax Calculation errors: 
screen shot 2018-08-10 at 11 56 04 am. @ghislaineguerin wondering if you can take a look.

Thanks.

ghislaineguerin:
@alex-deaconu How about this?
https://codepen.io/ghislaineguerin/pen/JBxoaW

alex-deaconu:
@ghislaineguerin thanks!
screen shot 2018-08-10 at 1 46 38 pm

alex-deaconu:
@olegrise I've updated the PR to implement the changes we discussed.

To confirm, I'm now hiding the Back button if the Form has already been completed (user clicks Edit from the Review Purchase page).

Also, the new Credit Card form will now be shown if the User clicks Edit for it, instead of the form with greyed out buttons.

Finally, new styling for the Error message and Retry buttons. See screenshot above.

@Rise-Vision/apps FYI. Please review. Thanks!

olegrise:
@alex-deaconu functionality looks good. I reviewed the code too, but not too thoroughly. 
@fjvallarino are review code as well?

src/js/components/purchase-flow/ftr-card-last-four.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+
+angular.module("risevision.common.components.purchase-flow")
+  .filter("cardLastFour", [
+
+    function () {
+      return function (last4) {
+        return "***-" + last4;
```
fjvallarino: @alex-deaconu just for the sake of safety, should this filter take the last four characters of the parameter it receives?

src/js/components/purchase-flow/svc-stripe.js
```diff
@@ -77,12 +77,12 @@ angular.module("risevision.common.components.purchase-flow")
 
         stripeLoader().then(function (stripeClient) {
           stripeClient.card.createToken(cardObject, function (status, response) {
-            if (response.error) {
-              card.tokenError = _processStripeError(response.error.code);
+            if (response && response.card && !response.error) {
+              deferred.resolve(response.card);
+            } else {
+              card.tokenError = _processStripeError(response.error && response.error.code);
```
fjvallarino: @alex-deaconu I assume this should not happen, but on the if clause you're checking ```response``` is not null. Should that check be done here too, or maybe removed from the if part?

fjvallarino:
@alex-deaconu looks good 👍 . Just two minor comments

alex-deaconu:
@fjvallarino addressed the two inline comments in the last commit. Please review. Thanks!

fjvallarino:
@alex-deaconu looks good! 👍

PR #860, opened 2018-08-14
Complete checkout process by calling purchase API

alex-deaconu:


alex-deaconu:
Contains the final step of the checkout process, as well as various fixes.

You can test the purchase flow with any company, but the payment step will not succeed. You should get an error similar to "Billing company must be migrated" if your Company is not origin="Chargebee".

If you do test with a migrated Company, you can replicate various Card errors with the Stripe test CC numbers (https://stripe.com/docs/testing#cards-responses).

Finally, you can successfully pay with the 4242 CC number. To confirm, your subscription will not refresh automatically when your transaction is completed (you need to reload the page). You can cancel the subscription (in ChargeBee, to test subscribing again), but it's not recommended you delete the subscription.

src/js/components/purchase-flow/services/svc-purchase-factory.js
```diff
@@ -155,6 +155,67 @@
...
+          //clean up items
+          var newItems = [{
+            id: _getChargebeePlanId()
+          }, {
+            id: _getChargebeeAddonId(),
+            qty: factory.purchase.plan.additionalDisplayLicenses
+          }];
+
+          var billTo = _copyAddress(factory.purchase.billingAddress);
+          billTo.id = factory.purchase.billingAddress.id;
```
fjvallarino: @alex-deaconu should the ```id``` field be also copied as part of _copyAddress?

src/js/components/purchase-flow/services/svc-purchase-factory.js
```diff
@@ -97,8 +97,8 @@
                 return stripeService.createToken(paymentMethods.newCreditCard, address)
                   .then(function (response) {
                     paymentMethods.newCreditCard.id = response.id;
```
fjvallarino: @alex-deaconu does ```id``` come as part of ```response```, or should it go through ```card``` too?

alex-deaconu:
@Rise-Vision/apps just a quick update to the PR, based on the latest commit. 

It is now using the real origin check from userState, so the payment form will only show for Companies that have been migrated.

In addition, it will now use the real shipping address fields from the (Selected) Company object.

@fjvallarino it also addresses the concern re: id field you brought up in your comment.

Please review. Thanks!

src/js/components/purchase-flow/services/svc-purchase-factory.js
alex-deaconu: @fjvallarino good question. This was an actual issue that was discovered. The Stripe createToken response has 2 IDs, one attached to the response object, and one on the response.card object. What we need to use is the one on the response object, which is why this change was required. Otherwise adding the card to ChargeBee fails.

This is also reflected in the Unit test here: https://github.com/Rise-Vision/common-header/pull/860/files#diff-17309909778e1b7ef4ce5b2757886b9fR65.

Thanks.

fjvallarino:
@alex-deaconu looks good 👍

olegrise:
@alex-deaconu for some reason Company Name field  is not populating for me neither on Billing Address page nor on Shipping Address.

alex-deaconu:
@olegrise good catch. I missed copying the company name fields with the addresses.

Please see last commit. Thanks.

olegrise:
@alex-deaconu looks good now 👍

PR #861, opened 2018-08-17
Make billingContactsEmails field editable

fjvallarino:
Can be tested with: https://apps-stage-1.risevision.com/?cid=a847f5bb-e61a-4925-817a-552075edf203, after opening Company Settings modal.


src/templates/company-fields.html
```diff
@@ -121,7 +121,18 @@
     <option value="{{c[1]}}" ng-repeat="c in timezones">{{c[0]}}</option>
   </select>
 </div>
-<div class="form-group" style="display: none">
+<div class="form-group" ng-show="company.isChargebee">
   <label class="form-control-label">Billing Notifications Email</label>
-  <input type="text" class="form-control" name="" value="user@companyabc.com">
+  <tags-input name="billingContactEmails"
```
olegrise: @fjvallarino I believe user should be able to enter multiple emails here, not one. @alex-deaconu

gulpfile.js
```diff
@@ -102,6 +103,7 @@ var env = process.env.NODE_ENV || "dev",
     "./bower_components/ng-biscuit/dist/ng-biscuit.js",
     "./bower_components/lodash/dist/lodash.js",
     "./bower_components/ng-csv/build/ng-csv.js",
+    "./bower_components/ng-tags-input/ng-tags-input.js",
```
alex-deaconu: @fjvallarino since this is being added here, it should probably be removed from index.html in Apps. Thanks.

src/js/controllers/ctr-company-settings-modal.js
```diff
@@ -1,16 +1,20 @@
 angular.module("risevision.common.header")
 
+.value("EMAIL_REGEX",
+  /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
```
alex-deaconu: @fjvallarino we already have this constant here https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/displays/controllers/ctr-display-details.js#L5. We should not duplicate this. Thanks.

src/templates/company-fields.html
alex-deaconu: @olegrise that is actually what that component does: 
screen shot 2018-08-20 at 10 40 05 am
Thanks.

src/js/controllers/ctr-company-settings-modal.js
fjvallarino: @alex-deaconu I know. I added it here because it was causing e2e tests failures (it took me a while to figure out what was causing them). We should remove the one in Apps.

src/templates/company-fields.html
olegrise: @alex-deaconu @fjvallarino awesome. Should the label be changed to "Billing Notifications Email**s**" i.e. plural form?

src/js/controllers/ctr-company-settings-modal.js
alex-deaconu: @fjvallarino makes sense. I would make a component for handling these multiple emails, that handles the validation as well as the conversion to and from list. Thanks.

alex-deaconu:
@olegrise doesn't Chargebee has a limit for the number of emails it can support for this field. Should we implement some sort of validation for this? @fjvallarino FYI. Thanks.

olegrise:
@alex-deaconu good question. Chargebee says you can have up to 5 contacts, but then @alexey-rise tested it and was able to add more than that from the API. Maybe only 5 will be used for sending emails? I don't know. Just to be safe it would be great if we could limit it to 5 in our UI. @fjvallarino

ezequielc:
@alex-deaconu @olegrise I just tested adding 10 emails and they were all added to Chargebee. I think that restricting to 5 will add complexity to the UI and validation. So, since we are short on time I would leave as is for now. If it becomes a problem by not sending the emails we can set a limit. 

But, @fjvallarino one thing that is missing is to list the emails on /billing page in the Account Information area. Thanks!

olegrise:
Sounds good re. keep it simple for now.

fjvallarino:
@Rise-Vision/apps updated

test/unit/controllers/ctr-company-settings-spec.js
```diff
@@ -5,6 +5,7 @@
 describe("controller: company settings", function() {
   beforeEach(module("risevision.common.header"));
   beforeEach(module(function ($provide, $translateProvider) {
+    $provide.value("EMAIL_REGEX", "");
```
alex-deaconu: @fjvallarino is this needed anymore? Thanks.

src/js/directives/dtv-emails-field.js
```diff
@@ -0,0 +1,81 @@
...
+      return {
+        restrict: "E",
+        require: "ngModel",
+        scope: {
+          emails: "=ngModel"
+        },
+        template: $templateCache.get("emails-field.html"),
+        link: function ($scope, elem, attr, ngModel) {
+          var EMAIL_REGEX =
+            /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
```
alex-deaconu: @fjvallarino I think we should keep this as a constant. I can imagine we can maybe re-use at some point. Thanks.

fjvallarino:
@alex-deaconu updated!

alex-deaconu:
@fjvallarino looks good 👍 . Thanks.

PR #862, opened 2018-08-20
Update Address and Contact info in Core

alex-deaconu:


alex-deaconu:
Code will check and update Addresses/Contact only if the values have actually changed.

The Updates are performed in the background (asynchronously) so the page will navigate while that is going on. Because of this, and the fact that these updates are not related to the Checkout process, no error message is shown in case of failure. Let me know if you have any thoughts on this behaviour.

src/js/components/purchase-flow/services/svc-contact.js
```diff
@@ -0,0 +1,40 @@
...
+        return (
+          c1.firstName === c2.firstName &&
+          c1.lastName === c2.lastName &&
+          c1.email === c2.email &&
+          c1.telephone === c2.telephone);
+      };
+
+      this.cleanContactObj = function (c) {
+        return {
+          username: c.username,
```
fjvallarino: @alex-deaconu this field is not needed by either ```contactsAreIdentical``` or ```copyContactObj```, right?

alex-deaconu:
@fjvallarino removed duplicate functionality, as per my comment. See last commit. Thanks.

fjvallarino:
@alex-deaconu looks good! 👍

PR #863, opened 2018-08-21
accessibility issues

ghislaineguerin:
@alex-deaconu Here are the accessibility updates. I will stage on apps, but I'm not sure which stage or if it's better to have you stage the changes. I do think we need at least 3 labels I've added to be included in the i18n files. Let me know or let's talk after standup.


src/templates/components/purchase-flow/checkout-review-subscription.html
```diff
@@ -48,7 +48,8 @@ <h3 class="text-center" translate translate-values="{ planName: plan.name }">com
               <i class="fa fa-caret-down"></i>
             </button>
           </div>
-          <input id="additionalLicenses" name="additionalLicenses" type="number" class="form-control" ng-model="plan.additionalDisplayLicenses" min="0" max="999" pattern="[0-9]{1,3}" required>
+          <label for="additionalLicenses" translate>common-header.purchase.review-subscription.additional-licenses</label>
+          <input id="additionalLicenses" name="additionalLicenses" type="number" class="form-control" ng-model="plan.additionalDisplayLicenses" min="0" max="999" pattern="[0-9]{1,3}" tabindex="1" aria-required="true" required>
```
alex-deaconu: @ghislaineguerin the label is appearing twice here:
screen shot 2018-08-21 at 12 34 23 pm
Can you remove one? Thanks.

alex-deaconu:
@ghislaineguerin as per our discussion, I've added a few fixes and updates. Wondering if you could please review the last commit. Thanks!

ghislaineguerin:
👍 @alex-deaconu

PR #865, opened 2018-08-22
update Twitter screenName validation error

santiagonoguez:
Updating the validation error message as suggested here:

https://github.com/Rise-Vision/rise-vision-apps/pull/719


sheadarlison:
@santiagonoguez @jungeunyoon Should the error message say handler or handle? And I don't think configuration means anything to a user.

Can we update this message to say: 

Error: you must specify a Twitter handle before you can save.

PR #868, opened 2018-08-24
Show Customer Portal when clicking Subscribe button for Chargebee users

fjvallarino:


src/js/components/plans/ctr-plans-modal.js
```diff
@@ -91,8 +91,10 @@ angular.module("risevision.common.components.plans")
           // Is lower Plan?
           if ($scope.currentPlan.order > plan.order) {
             return "downgrade";
```
alex-deaconu: @fjvallarino should we not be applying the same logic to Downgrade? Thanks.

fjvallarino: @alex-deaconu we always show Customer Portal for Downgrade, since the user already has a Chargebee account at that point. What do you suggest we should show?

alex-deaconu: @fjvallarino if they are on a Trial, we can show the Purchase Flow to be consistent. @ezequielc FYI. Thoughts? Thanks.

ezequielc: @alex-deaconu @fjvallarino Good point, if they are on trial we should show Purchase Flow. Thanks!

fjvallarino: @alex-deaconu @ezequielc updated. It will show the Purchase Flow using the plan the user has currently on Trial (which means, all the Downgrade buttons will show the same information)

src/templates/components/plans/plans-modal.html
```diff
@@ -61,19 +61,25 @@ <h3>{{plan.proLicenseCount}}</h3>
...
-          <a id="downgrade-plan" ng-show="getVisibleAction(plan) === 'downgrade' && !isChargebee()"
+          <a id="downgrade-plan" ng-if="!isChargebee()" ng-show="getVisibleAction(plan).indexOf('downgrade') === 0"
               target="_blank" href="https://www.risevision.com/downgradeplan"
               class="btn btn-default btn-block" translate>common-header.plans.downgrade</a>
-          <a id="downgrade-plan-cb" ng-show="getVisibleAction(plan) === 'downgrade' && isChargebee()"
+          <a id="downgrade-plan-cb" ng-if="isChargebee()" ng-show="getVisibleAction(plan) === 'downgrade'"
               ng-click="downgradePlan(plan)"
               class="btn btn-default btn-block" translate>common-header.plans.downgrade</a>
+          <a id="downgrade-plan-cbt" ng-if="isChargebee()" ng-show="getVisibleAction(plan) === 'downgrade-trial'"
+              ng-click="purchaseFactory.showPurchaseModal(currentPlan, isMonthly)"
```
alex-deaconu: @fjvallarino I think `showPurchaseModal` should be called with `plan` here. Thanks.

src/js/components/plans/ctr-plans-modal.js
```diff
@@ -90,9 +90,15 @@ angular.module("risevision.common.components.plans")
         } else { // This is a different Plan
           // Is lower Plan?
           if ($scope.currentPlan.order > plan.order) {
-            return "downgrade";
-          } else { // Higher Plan
+            if (currentPlanFactory.isOnTrial()) {
+              return "downgrade-trial";
```
alex-deaconu: @fjvallarino I think you should make these names more consistent between each case.
`downgrade-trial` - `subscribe`
`downgrade` - `subscribe-portal`.

I would suggest `downgrade` and `downgrade-portal`. Thanks.

src/templates/components/plans/plans-modal.html
fjvallarino: @alex-deaconu I don't think that's a good options, because otherwise it would show a $0 plan when clicking on Free. That's the reason I went with ```currentPlan```

alex-deaconu: @fjvallarino I don't follow. If I have a trial for the Basic plan, and I click Downgrade to the Starter, in your code the Purchase Modal would open with the Basic Plan selected (instead of Starter). We probably need to add a condition for Free in this case.

fjvallarino: @alex-deaconu well, I think any of the options is actually confusing. Clicking Downgrade while On Trial and being shown the Purchase Flow sounds weird; I'm not sure it's clear for users it will cancel their current Trial and Subscribe them to a lower plan (maybe they expect they can start a trial on a lower plan?). Anyway, let's handle it as you propose. I will show Customer Portal for Free plan.

ezequielc: @fjvallarino I think we should actually show Subscribe instead of Downgrade in this case. But this is not a priority and we can improve later. We need to address the current issues so we can provide a stage version to Support today. Thanks

fjvallarino:
@alex-deaconu updated!

src/templates/components/plans/plans-modal.html
```diff
@@ -50,30 +50,55 @@ <h3>{{plan.proLicenseCount}}</h3>
...
           <p class="small u_margin-sm-bottom text-danger" ng-show="isCurrentPlan(plan) && isTrialExpired(plan)" translate>common-header.plans.trial-expired</p>
 
-          <a id="subscribe-plan" ng-if="!isChargebee()" ng-show="getVisibleAction(plan) === 'subscribe'"
+          <a id="subscribe-plan" ng-if="!isChargebee()" ng-show="getVisibleAction(plan).indexOf('subscribe') === 0"
               target="_blank" href="https://store.risevision.com/product/{{plan.productId}}"
               class="btn btn-primary btn-block" translate>common-header.plans.subscribe</a>
-          <a id="subscribe-plan" ng-if="isChargebee()" ng-show="getVisibleAction(plan) === 'subscribe'"
+<!-- Remove after discussion with Chargebee gives us a definitive solution -->
+          <a id="subscribe-plan-cb" ng-if="isChargebee()" ng-show="getVisibleAction(plan).indexOf('subscribe') === 0"
+              ng-click="purchaseFactory.showPurchaseModal(plan, isMonthly)"
```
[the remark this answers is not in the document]
fjvallarino: @ezequielc @alex-deaconu sorry, what the url for that form is? We currently have:

- Additional Licenses: https://www.risevision.com/purchaseadditionaldisplaylicenses
- Downgrade: https://www.risevision.com/downgradeplan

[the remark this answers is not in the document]
ezequielc: @fjvallarino I am checking with JR and Scott, but for now you can use the same as Downgrade, I will work on updating the wording. Thanks

alex-deaconu:
@fjvallarino looks good 👍

PR #874, opened 2018-08-30
Use extend instead of copy

alex-deaconu:
copy clears out all other properties from the dest object


alex-deaconu:
Missed this as per angular.copy() documentation:
> If a destination is provided, all of its elements (for arrays) or properties (for objects) are deleted and then all elements/properties from the source are copied to it.

Using angular.extend() instead (https://docs.angularjs.org/api/ng/function/angular.extend).

PR #877, opened 2018-09-03
Track Store events

fjvallarino:


src/js/components/purchase-flow/controllers/ctr-purchase-modal.js
```diff
@@ -37,11 +37,51 @@ angular.module("risevision.common.components.purchase-flow")
...
+    }
+
+    function _trackOrderPayNowClicked() {
+      var estimate = purchaseFactory.purchase.estimate || {};
+
+      if (!estimate.estimateError) {
+        segmentAnalytics.track("Order Pay Now Clicked", {
+          amount: estimate.total,
+          currency: estimate.currency,
+          inApp: userState.inRVAFrame()
```
alex-deaconu: @fjvallarino I don't think this can ever be `true`, so we can hardcode or omit. Thanks.

src/js/components/purchase-flow/controllers/ctr-purchase-modal.js
```diff
@@ -37,11 +37,51 @@ angular.module("risevision.common.components.purchase-flow")
     $scope.$watch("factory.loading", function (loading) {
       if (loading) {
         $loading.start("purchase-modal");
+        _trackProductAdded();
```
alex-deaconu: @fjvallarino the spinner is called at various points of the checkout process, so this event would be tracked multiple times for the same checkout. If the goal is tracking when the modal is opened, then we should track in `purchaseFactory.showPurchaseModal`. Thanks.

src/js/components/purchase-flow/controllers/ctr-purchase-modal.js
```diff
@@ -98,7 +139,10 @@ angular.module("risevision.common.components.purchase-flow")
 
         $scope.finalStep = true;
 
-        purchaseFactory.getEstimate();
+        purchaseFactory.getEstimate()
+          .then(function () {
+            _trackPlaceOrderClicked();
```
alex-deaconu: @fjvallarino why are we tracking after the result? Should we not track on click? Thanks.

alex-deaconu:
@fjvallarino as a general note, I would prefer if the Tracking code was split in its own factory, and if the tracking was triggered in the purchaseFactory rather than in the controller. Thanks.

alex-deaconu:
@fjvallarino the main advantage I see to creating a separate factory is less code in the original factory. Also, it would prevent some code duplication if we were to call events from various places. For example controller and purchaseFactory, or split purchaseFactory into multiple factories (which I'm actually considering). Thanks.

fjvallarino:
@alex-deaconu updated!

src/js/components/purchase-flow/services/svc-track-events.js
```diff
@@ -0,0 +1,41 @@
...
+          name: plan.name,
+          price: plan.isMonthly ? plan.monthly.billAmount : plan.yearly.billAmount,
+          quantity: 1,
+          category: "Plans",
+          inApp: false
+        });
+      };
+
+      factory.trackPlaceOrderClicked = function (estimate) {
+        if (!estimate.estimateError) {
```
alex-deaconu: @fjvallarino I don't think we need to check for the estimateError anymore. We only track on the success case for this. Thanks.

src/js/components/purchase-flow/services/svc-track-events.js
```diff
@@ -0,0 +1,41 @@
...
+          segmentAnalytics.track("Place Order Clicked", {
+            amount: estimate.total,
+            currency: estimate.currency,
+            inApp: false
+          });
+        }
+      };
+
+      factory.trackOrderPayNowClicked = function (estimate) {
+        if (!estimate.estimateError) {
```
alex-deaconu: @fjvallarino same as above, this can be removed. I don't think we need to check for errors (plus this is the incorrect error object to check, it should be checkoutError). Thanks.

src/js/components/purchase-flow/services/svc-track-events.js
```diff
@@ -0,0 +1,41 @@
+"use strict";
+
+angular.module("risevision.common.components.purchase-flow")
+  .factory("trackEvents", ["segmentAnalytics",
```
alex-deaconu: @fjvallarino I would rename this to something more specific, and consistent with other tracker services from Apps (`scheduleTracker`, etc), like `purchaseFlowTracker`. Thanks.

alex-deaconu:
@fjvallarino looks good, just a few additional comments for cleanup. Thanks.

fjvallarino:
@alex-deaconu updated again

alex-deaconu:
@fjvallarino looks good, thanks for fixing 👍

Re version number, I know you're expecting another PR to be merged first. However, if you're planning to merge this first, don't forget to bump down the version.

Thanks.

PR #879, opened 2018-09-06
Add Tax Exemption

fjvallarino:
I'm on the fence about creating a separate directive for the File Upload component. I don't really like handling that code in the controller, but I'm also not sure if it's worth spending more time on this. Let me know your thoughts!


fjvallarino:
@alex-deaconu updated!

alex-deaconu:
@fjvallarino couple of issues I found while testing this form:
- When I click Attach File, the whole form goes through validation and shows the fields as invalid
- The form needs to be accessible, maybe you can work with @ghislaineguerin on updating
- The file field's left side is not rounded off, it looks inconsistent with the other fields in the form
- The "Exemption State/Province *" text is showing even if the field is not there, with a validation failure

Also, I don't think we need to show the "Missing ... field" error messages for the form; we don't have that pattern anywhere. I believe the Asterisks and fields being highlighted are clear indicators that values are missing.

src/js/components/purchase-flow/controllers/ctr-tax-exemption-modal.js
```diff
@@ -0,0 +1,118 @@
...
+          errors.push("Missing Exempt in State");
+        }
+
+        return errors;
+      };
+
+      $scope.selectFile = function () {
+        setTimeout(function () {
+          document.querySelector("#inputExemption").click();
+        }, 0);
```
alex-deaconu: @fjvallarino why is this timeout needed? Thanks.

src/js/components/purchase-flow/controllers/ctr-tax-exemption-modal.js
```diff
@@ -0,0 +1,118 @@
...
+      };
+
+      $scope.setFile = function (element) {
+        $scope.$apply(function () {
+          $scope.fileNotSelected = true;
```
alex-deaconu: @fjvallarino what is this variable used for? Thanks.

src/js/components/purchase-flow/controllers/ctr-tax-exemption-modal.js
```diff
@@ -0,0 +1,118 @@
...
+        var field = form[fieldName];
+
+        return (field.$dirty || form.$submitted) && !_getProvinceValue();
+      };
+
+      function _getProvinceValue() {
+        if ($scope.formData.country === "US") {
+          return $scope.formData.state;
+        } else {
+          return $scope.formData.province;
```
alex-deaconu: @fjvallarino why not use the same ng-model value for the province? It works for all the other forms, and you can use the required attribute in that case. Thanks.

src/js/components/purchase-flow/controllers/ctr-tax-exemption-modal.js
fjvallarino: @alex-deaconu to avoid a $digest in progress exception

src/js/components/purchase-flow/controllers/ctr-tax-exemption-modal.js
fjvallarino: @alex-deaconu removed

src/js/components/purchase-flow/controllers/ctr-tax-exemption-modal.js
fjvallarino: @alex-deaconu I'll try to copy it again from the existing forms. Previously I was having issues with error label not being updated correctly

alex-deaconu: @fjvallarino did you try it without the text field in question? I think it should still work. Either way, I'm good with leaving as is; just extra code. Thanks.

fjvallarino: @alex-deaconu the main issue I see at this point is I would have to change the way the validation works (currently, it uses ```isFieldInvalid('province')```, which means I need a field in the form whose name is ```province```). I was having issues earlier where that validation would not update correctly, which is why I settled on using a hidden field (simulating the real input you have in your form)

alex-deaconu:
@fjvallarino looks good 👍 . Thanks for the fixes.

Let's go with the PR as is and have @ghislaineguerin address the accessibility in a separate card. @ezequielc let me know if that works for you. Thanks.

src/js/components/purchase-flow/controllers/ctr-tax-exemption-modal.js
alex-deaconu: @fjvallarino okay, let's leave as is. Thanks.

fjvallarino:
@alex-deaconu sure. What the accessibility issues? I tried the mobile version using Chrome's DevTools and it looks _decent_ now.

alex-deaconu:
@fjvallarino basically using the `for="field"` attribute for the labels and setting the tabindex for the fields, etc. Thanks.

fjvallarino:
@alex-deaconu @ezequielc is this good to merge?

alex-deaconu:
@fjvallarino looks good to me 👍 . Thanks.

PR #883, opened 2018-09-16
Redirect Chargebee to Apps

fjvallarino:
Since ```STORE_URL``` is redefined in the Store and its value replaced with "", I created a copy named ```STORE_URL_CH```. I use its value to check if the user is currently in Apps or Store, and what the current path is (I could use routes for that part). Let me know if you have a better idea on how to implement this (or a better name for the new constant).


src/js/dtv-common-header.js
```diff
@@ -167,6 +166,17 @@ angular.module("risevision.common.header", [
         event.preventDefault();
       }
     });
+
+    $rootScope.$on("risevision.company.selectedCompanyChanged", function () {
+      // Redirect Store requests by Chargebee users to Apps
+      var currentURL = $window.location.href;
+      var isStoreRequest = currentURL.indexOf(STORE_URL_CH) >= 0;
+      var isInvoicesRequest = currentURL.indexOf(STORE_URL_CH + "account/view/invoicesHistory") >= 0;
```
alex-deaconu: @fjvallarino I don't follow why this check is in the Common-Header? We should move this to the Store as it specific to that App. Thanks.

fjvallarino:
@alex-deaconu moved redirection logic to Store and removed unneeded constants.

alex-deaconu:
@fjvallarino looks good 👍 

Just curious re: unit test changes, does that fix the randomly failing unit test we've seen before?

Thanks!

fjvallarino:
@alex-deaconu the useFakeTimers thing fixes this issue:

image

alex-deaconu:
Nice!!

olegrise:
👍

PR #889, opened 2018-10-12
Allow editing shareCompanyPlan

fjvallarino:


src/js/controllers/ctr-company-settings-modal.js
```diff
@@ -43,6 +43,7 @@ angular.module("risevision.common.header")
           $scope.company = company;
           $scope.company.isSeller = company && company.sellerId ? true : false;
           $scope.company.isChargebee = company && company.origin === "Chargebee";
+          $scope.company.ownsPlan = company && ["Active", "Trial"].indexOf(company.planSubscriptionStatus) >= 0;
```
alex-deaconu: @fjvallarino why do we have this extra condition? I don't think this is necessary and we can show the checkbox for any Company regardless of the Plan status. Thanks.

alex-deaconu:
@fjvallarino one comment inline. Don't forget to bump the version before merging. Thanks.

src/templates/company-fields.html
```diff
@@ -121,6 +121,10 @@
     <option value="{{c[1]}}" ng-repeat="c in timezones">{{c[0]}}</option>
   </select>
 </div>
+<div class="form-group" require-role="sa">
```
alex-deaconu: @fjvallarino as per discussion, we should probably also update this require-role="ba". Thanks.

src/js/controllers/ctr-company-settings-modal.js
```diff
@@ -156,7 +156,6 @@ angular.module("risevision.common.header")
         //exclude fields from API call
         delete company.sellerId;
         delete company.isTest;
-        delete company.shareCompanyPlan;
```
alex-deaconu: @fjvallarino isn't this still necessary for other roles than RiseStoreAdmin? Thanks.

fjvallarino: @alex-deaconu it actually is; I will restore it. I'm currently testing a couple of things. I'll let you know as soon as the PR is ready for review

fjvallarino:
@alex-deaconu @olegrise this is ready for review again

src/js/controllers/ctr-company-settings-modal.js
```diff
@@ -156,6 +156,7 @@ angular.module("risevision.common.header")
         //exclude fields from API call
         delete company.sellerId;
         delete company.isTest;
+        delete company.shareCompanyPlan;
```
alex-deaconu: @fjvallarino just a note, I had no idea this was a thing. Good pattern to keep in mind for other APIs, especially if we want to fix the Presentation API issue with the template sharing field. Thanks.

alex-deaconu:
@fjvallarino looks good 👍 , thanks for fixing!

PR #892, opened 2018-10-18
Fix Plan Banner use case for Cancelled Active subscriptions

fjvallarino:


src/js/components/plans/svc-current-plan-factory.js
```diff
@@ -16,6 +16,7 @@
             plan = _.cloneDeep(_plansByCode[company.planProductCode]);
             plan.status = company.planSubscriptionStatus;
             plan.trialPeriod = company.planTrialPeriod;
+            plan.currentPeriodEndDate = new Date(company.planCurrentPeriodEndDate);
```
fjvallarino: @alex-deaconu could ```company.planCurrentPeriodEndDate``` be null?

alex-deaconu: @fjvallarino yes, it is in most cases. We just get an invalid date in that case.

PR #894, opened 2018-10-30
Fix issue with stripping leading 0 from CVC

alex-deaconu:
Staged at https://apps-stage-0.risevision.com


src/templates/components/purchase-flow/checkout-payment-methods.html
```diff
@@ -90,7 +90,7 @@
           <div class="col-md-4">
             <div class="form-group" ng-class="{ 'has-error': (form.paymentMethodsForm.cardCvc.$dirty || form.paymentMethodsForm.$submitted) && form.paymentMethodsForm.cardCvc.$invalid }">
               <label for="new-card-cvc" class="control-label">Security Code *</label>
-              <input id="new-card-cvc" aria-required="true" tabindex="1" type="number" class="form-control" name="cardCvc" data-stripe="cvc" ng-model="paymentMethods.newCreditCard.cvc" autocomplete="cc-csc" maxlength="4" required>
+              <input id="new-card-cvc" aria-required="true" tabindex="1" type="text" pattern="[0-9]*" class="form-control" name="cardCvc" data-stripe="cvc" ng-model="paymentMethods.newCreditCard.cvc" autocomplete="cc-csc" maxlength="4" required>
```
fjvallarino: @alex-deaconu should we restrict this to three characters?

alex-deaconu: @fjvallarino actually 4 characters, and it is restricted via `maxlength="4"` near the end of the tag. Thanks.

fjvallarino: @alex-deaconu nice, I didn't check the end of tag!

PR #896, opened 2018-11-15
Update Account & Billing link to always show /billing for Chargebee users

fjvallarino:


fjvallarino:
This can be tested on stage-1 (https://apps-stage-1.risevision.com/?cid=c09e1aff-93f2-4c50-b964-44eafb6f75b2)

src/js/directives/dtv-website-validator.js
```diff
@@ -9,7 +9,7 @@ angular.module("risevision.common.header.directives")
         restrict: "A",
         link: function (scope, elem, attr, ngModel) {
           var WEBSITE_REGEXP =
-            /^(http[s]?:\/\/){0,1}([^\s/?\.#:@"]+\.)+([^\s/?\.#:"@-]{2,5})([\/?#][^\s"]*)?$/;
+            /^(http[s]?:\/\/){0,1}([^\s/?\.#:@"]+\.)+([^\s/?\.#:"@-]{2,61})([\/?#][^\s"]*)?$/;
```
alex-deaconu: @fjvallarino is `61` intentional here, or just `6`? Thanks.

src/js/controllers/ctr-auth-buttons.js
```diff
@@ -1,13 +1,14 @@
 angular.module("risevision.common.header")
+  .value("RVA_BASE_URL", /https*:\/\/rva(-test)*.risevision.com/)
```
alex-deaconu: @fjvallarino there's no rva-test.risevision.com, instead it's rva-test.appspot.com. 

I think you should check for rva.risevision.com (ignoring .appspot.com domains), but also I would allow both http and https, as some people may use RVA with http.

Thanks.

src/js/controllers/ctr-auth-buttons.js
```diff
@@ -1,13 +1,14 @@
...
+    "$loading", "cookieStore",
     "$log", "uiFlowManager", "oauth2APILoader", "bindToScopeWithWatch",
-    "$window", "$state", "APPS_URL",
+    "$window", "APPS_URL", "RVA_BASE_URL",
     function ($scope, $modal, $templateCache, userState, userAuthFactory,
       canAccessApps,
-      $loading, cookieStore, plansFactory, currentPlanFactory, $log, uiFlowManager, oauth2APILoader,
-      bindToScopeWithWatch, $window, $state, APPS_URL) {
+      $loading, cookieStore, $log, uiFlowManager, oauth2APILoader,
+      bindToScopeWithWatch, $window, APPS_URL, RVA_BASE_URL) {
```
alex-deaconu: @fjvallarino what is APPS_URL needed for here? Thanks.

src/templates/auth-buttons-menu.html
```diff
@@ -23,7 +23,11 @@
       <i class="fa fa-shopping-cart"></i>
       <span class="item-name">Account & Billing</span>
     </a>
-    <a href="#" class="store-account-button action" ng-click="showAccountAndBilling()" ng-show="isChargebee()">
+    <a ui-sref="apps.billing.home" class="store-account-button action" ng-show="isChargebee() && !isRVA()">
```
alex-deaconu: @fjvallarino if we open in a new tab for RVA, what about Store, Developer Hub and any other potential places CH may be used for? I think maybe we should check if the page `isApps`. Thoughts? Thanks.

src/js/controllers/ctr-auth-buttons.js
fjvallarino: @alex-deaconu it's user for the Alerts settings URL

src/js/controllers/ctr-auth-buttons.js
fjvallarino: @alex-deaconu it does check for http/https (the _s*_ part)

src/templates/auth-buttons-menu.html
fjvallarino: @alex-deaconu makes sense, yes

src/js/directives/dtv-website-validator.js
fjvallarino: @alex-deaconu it's intentional. That's the maximum a domain label can have. I guess we could use a lower number, though

alex-deaconu: @fjvallarino oh, I didn't know. I'm good to leave as is in that case. Thanks.

alex-deaconu:
@fjvallarino looks good 👍 . Thanks for fixing!

PR #898, opened 2018-11-20
 Remove usage of messagesPrefix in subscription-status

stulees:
Fixes widget issue https://github.com/Rise-Vision/widget-video/issues/251


alex-deaconu:
To confirm, I've double checked both common-header and rise-vision-apps and that variable does not seem to be used in the SubscriptionStatus Component.

@stulees looks good 👍 . Thanks for addressing!

stulees:
@fjvallarino Okay I'll try again

PR #907, opened 2019-01-18
Fix issue with Timeline Date intialization

alex-deaconu:
Handle useLocaldate usecase


alex-deaconu:
Fix for https://github.com/Rise-Vision/rise-vision-apps/issues/796

There were a few issues to contend with when implementing this fix. First off, the Timeline `startDate`/`endDate` are actually String values. We use a directive to modify the values for the `ng-model` of the DatePicker. This was resolved by parsing the Dates.

However, the Timeline component also supports dates in 2 different formats (Java and Javascript Dates, with and without timezone). This also needed to be taken into account, based on the flag set in the directive.

Similar to this code: https://github.com/Rise-Vision/common-header/blob/master/src/js/components/timeline/dtv-datepicker-localdate.js#L44.

You can validate here: https://apps-stage-3.risevision.com/schedules/details/bdec4dc9-2279-4e11-a59e-344fb6066cca?cid=17899fe3-db05-4ecd-ade4-a7106fe53784. Schedule Timeline (Java Date) and Schedule Item Timelines (JS Date) use the 2 different date formats.

PR #912, opened 2019-02-18
Add host name, cpu, product name

tejohnso:


alex-deaconu:
@tejohnso I think you forgot to increment the version in `bower.json`/`package.json`. You can run `gulp bump` to increment minor versions in Common Header. Thanks!

PR #918, opened 2019-02-26
Handle Presentation deletion on Template Editor

fjvallarino:
@stulees @santiagonoguez please review


bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "common-header",
-  "version": "3.5.0",
+  "version": "3.5.1",
```
stulees: @fjvallarino You need to bump this again, I used 3.5.1 in my last PR which is merged. Thanks

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "common-header",
-  "version": "3.5.0",
+  "version": "3.5.1",
```
stulees: @fjvallarino Same here

bower.json
fjvallarino: @stulees yes, I had pushed a commit a few minutes ago. I'll push a new one merging master

PR #926, opened 2019-03-11
Feature/add nyse nasdaq labels

santiagonoguez:
Add labels for ftf-high-nyse-nasdaq template.


src/locales/locales_json/en/template.json
```diff
@@ -28,13 +28,15 @@
       "stocks": "Most Popular Stocks",
       "world-indexes": "Most Popular World Indexes"
     },
+    "nasdaq-stocks": "NASDAQ Stocks",
```
stulees: @santiagonoguez I'm wondering, is it a necessity to be adding financial instance labels to common-header? Are we setting a precedent that Creative will have to continuously add labels here if they don't already exist? What reason do we have against just setting _NASDAQ Stocks_ as the value to the label?

santiagonoguez: @stulees we designed as this because the Software Architecture Process requires localization for our software.

https://help.risevision.com/hc/en-us/articles/360000045486-Software-Architecture-Design-Process

stulees: @santiagonoguez Yes, I suppose I'm questioning the value in translating a component instance label, particularly in how this will also scale for other components we build and use in Templates. For example, what if an HTML template has 25 instances of components and Creative want to label each one differently - 25 new translations for 1 template. This is a discussion separate from this PR, I'll raise this with the team. Cheers

PR #936, opened 2019-03-27
Use 'and' instead of '&' for Industry

alex-deaconu:
Minor version bump


alex-deaconu:

PR #940, opened 2019-03-27
Refactor HTML forms

alex-deaconu:
Add accessibility
Fix checkboxes
Autofocus
Submit with enter key


alex-deaconu:
@olegrise as discussed, a few improvements for the HTML forms. Please review. Thanks!

PR #941, opened 2019-03-28
Populate industry from URL param

olegrise:
@Rise-Vision/apps I wonder if you have any ideas how to resolve it. Basically if I (a new user) open URL 

https://apps-stage-3.risevision.com/editor/workspace/new/?copyOf=787b9532-a46e-483d-8477-d71d3ed5a205&industry=Gyms%20and%20Fitness

I cannot read the `industry` parameter from URL because `$location.search()` is empty. However if I refresh the page, then `$location.search().industry` is there. I'm currently debugging CH trying to get a clear understanding of what is going on, but I wonder if you already know and maybe can give me a tip.


olegrise:
@Rise-Vision/apps please review. 

The solution for parsing URL is pretty ugly. I considered an option of moving the parsing part to the `urlStateService` and do a combination of `JSON.parse` and `decodeURIComponent`, but at the end it would still be hard to understand what is going on there and it would still use regex. 

My final decision was to keep the code as simple as possible i.e. just call `decodeURIComponent` multiple times and keep the parsing function in the controller because this use case is very specific to this modal.

You can use this URL for testing https://apps-stage-3.risevision.com/editor/workspace/new/?copyOf=787b9532-a46e-483d-8477-d71d3ed5a205&industry=Gyms%20and%20Fitness

alex-deaconu:
@olegrise looks good 👍 . Interesting approach on the mocks for `$location`. I didn't know it was possible to set values using the `$$path`, etc. Thanks!

PR #943, opened 2019-03-29
Fix issue with validation fields missing

alex-deaconu:
Use conditional required field validation


alex-deaconu:
Fixes issues with User registration. The fields were removed via `ng-if` and are not found in the controller, generating an error. Reversed to the original logic of using `ng-show` and `ng-required`.

alex-deaconu:
Staged at https://apps-stage-0.risevision.com. Thanks!

alex-deaconu:
@olegrise thanks for the quick review!

PR #946, opened 2019-04-02
Add e2e test for Existing Company User registration

alex-deaconu:
Flow:
Add a New User to Existing Company
Sign out of Current Account
Log in to the New User Account
Complete Registration
Delete the New User

Note:
Using jenkins@risevision.com and his Company (account is used for Apps e2e tests) to add the new User.

The new User added is jenkins1@risevision.com which is used for e2e tests.

To prevent incorrect User and Company deletion, I've added checks for the Name of the Company to be deleted, and the Email of the User being deleted.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -295,6 +295,8 @@ gulp.task("test:e2e:core", ["test:webdrive_update"], factory.testE2EAngular({
   browser: "chrome",
   loginUser: process.env.E2E_USER2,
   loginPass: process.env.E2E_PASS2,
+  loginUser2: process.env.E2E_USER,
+  loginPass2: process.env.E2E_PASS,
   testFiles: function(){ 
     try{
       return JSON.parse(fs.readFileSync('/tmp/testFiles.txt').toString())
--- a/test/e2e/cases/registration-existing-company.js
+++ b/test/e2e/cases/registration-existing-company.js
@@ -1,0 +1,183 @@
+"use strict";
+var expect = require("rv-common-e2e").expect;
+var helper = require("rv-common-e2e").helper;
+var CommonHeaderPage = require("rv-common-e2e").commonHeaderPage;
+var HomePage = require("./../pages/homepage.js");
+var RegistrationModalPage = require("./../pages/registrationModalPage.js");
+var CompanyUsersModalPage = require("./../pages/companyUsersModalPage.js");
+var UserSettingsModalPage = require("./../pages/userSettingsModalPage.js");
+var helper = require("rv-common-e2e").helper;
+
+var FirstSigninScenarios = function() {
+
+  browser.driver.manage().window().setSize(1400, 900);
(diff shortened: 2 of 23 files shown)


alex-deaconu:

gulpfile.js
```diff
@@ -295,6 +295,8 @@ gulp.task("test:e2e:core", ["test:webdrive_update"], factory.testE2EAngular({
   browser: "chrome",
   loginUser: process.env.E2E_USER2,
   loginPass: process.env.E2E_PASS2,
+  loginUser2: process.env.E2E_USER,
+  loginPass2: process.env.E2E_PASS,
```
olegrise: @alex-deaconu 
`loginUser`**2**`: E2E_USER`
`loginUser: E2E_USER`**2**
😉

alex-deaconu: @olegrise yup, that is correct. I'll move this discussion to the card as it deals with credentials. Thanks.

test/e2e/cases/registration-existing-company.js
```diff
@@ -0,0 +1,185 @@
...
+
+        it("opens up Add User dialog", function () {
+          companyUsersModalPage.getAddUserButton().click();
+          
+          helper.wait(userSettingsModalPage.getUserSettingsModal(), "User Settings Modal");
+
+          expect(userSettingsModalPage.getUserSettingsModal().isPresent()).to.eventually.be.true;
+        });
+
+        it("adds a user", function () {
```
olegrise: @alex-deaconu We already have "add a user" [scenario](https://github.com/Rise-Vision/common-header/blob/62760d2adf83f683dfc1e7113b469b09a1ef4c51/test/e2e/cases/company-users.js#L61) covered. I'm not 100% sure what is the best way to handle it, but it doesn't seem right that we repeat the same tests in different places. What I'm trying to say here is that I think it's okay to repeat the same steps to re-create situation, however it should be clear **what exactly** we are testing. For example "add a user" scenario should be tested in `company-users.js`. Then in the `registration-existing-company.js`, the "add a user" scenario should be provided by common code or mockups and the focus should be on the registration module. I'm not worried about duplication of the code so much, even though it would be nice if we could avoid it, as clarity of what is being tested. Right now it seems that `registration-existing-company.js` tests scenarios outside of its domain.

alex-deaconu:
@olegrise please review the latest commits. Hope you enjoy ;).
@ezequielc FYI.
Thanks!

olegrise:
@alex-deaconu looks good. Curious, what was the reason for switching to using get functions?

alex-deaconu:
@olegrise you mean from using get functions? The get functions are only available in an instance object, so I would have to do `var self = this;` to be able to use `self.get...()`. I opted for using the local variables instead. Thanks.

olegrise:
@alex-deaconu ha-ha, sorry, you are right. I somehow saw it in reversed order 🤪

PR #947, opened 2019-04-04
Show Plan Managed by Parent if purchased by Parent

alex-deaconu:
Update Plan bar wording, hide Change Plan CTA

Handle Chargebee 404 error for Company who has
a Plan purchased by Parent

Do not allow Plans Modal to open more than once

Update fields as per Core
Updated i18n strings


alex-deaconu:
Part of the fix for https://github.com/Rise-Vision/rise-vision-apps/issues/819.

alex-deaconu:
The Company/Display I used for testing is https://apps-stage-0.risevision.com/displays/details/4QZP6K66NERZ?cid=c98270b9-92e6-4ac0-8d55-bc47cd9add52.

You can toggle licenses on and off on the 2 displays to validate functionality.

Thanks.

src/js/services/svc-chargebee-factory.js
```diff
@@ -75,9 +76,13 @@ angular.module("risevision.store.services")
           });
       }
 
-      function _handleChargebeeAccountNotFound(err, companyId) {
-        if (err.status === 404) {
+      function _handleChargebeePortalError(err, companyId) {
+        if (err.status === 404 && !currentPlanFactory.currentPlan.isPurchasedByParent && !plansFactory.isPlansModalOpen) {
```
ezequielc: @alex-deaconu Isn't `!plansFactory.isPlansModalOpen` check redundant here? It is also performed inside `plansFactory.showPlansModal()`. Thanks

ezequielc:
@alex-deaconu When I am in the Billing page, and I switch to the ... (Parent) company, it shows the 'You don't have access to make changes to your Subscription. Please contact your Plan Administrator.' message. Maybe it is just a variable that needs to be cleaned, because I don't see that message when refreshing the page. Thanks!

src/js/services/svc-chargebee-factory.js
alex-deaconu: @ezequielc the purpose of this is to show the Error in the PlansModal if it still happens to occur. Thanks.

alex-deaconu:
@ezequielc good point, I will work on clearing the message. Thanks!

src/js/services/svc-chargebee-factory.js
ezequielc: @alex-deaconu Right, my suggestion was that you could remove `!plansFactory.isPlansModalOpen` from line 80 and the result would be the same. But maybe I am missing something. Thanks

alex-deaconu: @ezequielc to confirm, the flow for these `if` statements is the following:
- User clicks something, and gets 404 error, that results in the Plans Modal to be opened (without an error message).
- User clicks something in the Plans Modal, and gets a 404 error, that results in the error message to appear in the Plans Modal.

The second step should NOT happen, but if it does, this is what we want the flow to be. If my understanding of what you're suggesting is correct, if we remove line 80, we would show the Plans Modal with an Error message which is incorrect.

Based on the above, the redundant check is https://github.com/Rise-Vision/common-header/pull/947/files#diff-d268d853f7432a726682ffe372ed702bR120, but I added because I felt it's good insurance.

Thanks.

ezequielc: @alex-deaconu Sounds good. In this case am fine with it. Thanks!

src/templates/plan-banner.html
```diff
@@ -3,13 +3,20 @@
...
+        <!-- <a href="#" aria-label="Go to Parent Company">{{plan.parentPlanCompanyName}}</a> -->
       </span>
       <a href="#" ng-click="showPlans()" class="ml-3 btn btn-xs btn-primary">Change Plan</a>
     </div>
   </div>
 
+  <div id="parent-purchased-plan-banner" class="alert p-2 mb-0 bg-success text-right" ng-show="getVisibleBanner() === 'parentPurchased'">
+    <div class="u_margin-right">
+      <span class="font-weight-bold">{{plan.name}} Plan</span>
+      <span class="ml-2">Managed by {{plan.parentPlanCompanyName || "the Parent Company"}}
```
olegrise: @alex-deaconu it looks like `<span>` tag is not closed.

src/templates/plan-banner.html
```diff
@@ -3,13 +3,20 @@
   <div id="parent-plan-banner" class="alert p-2 mb-0 bg-success text-right" ng-show="getVisibleBanner() === 'parent'">
     <div class="u_margin-right">
       <span class="font-weight-bold">{{plan.parentPlan.name}} Plan</span>
-      <span class="ml-2">Managed by {{plan.parentPlan.companyName || "the Parent Company"}}
-        <!-- <a href="#" aria-label="Go to Parent Company">{{plan.parentPlan.companyName}}</a> -->
+      <span class="ml-2">Managed by {{plan.parentPlanCompanyName || "the Parent Company"}}
```
olegrise: @alex-deaconu I see a lot of `<span>` tags that are formatted in a very strange way like this 
```
<span ... </span>
```
i.e. the closing `>` is missing.

olegrise: Found the missing `>`. Disregard my comment :)

alex-deaconu:
@ezequielc while I'm still waiting for CCI to finish the build/tests, this is now staged at https://apps-stage-0.risevision.com.

To confirm, the issue you brought up with the message not clearing from the billing page should now be fixed. 

@olegrise FYI. Wondering if you could review the latest commits.

Thanks!

ezequielc:
@alex-deaconu Looks good. Thanks for fixing it!

PR #949, opened 2019-04-11
Add name parameter to selectSubCompany

fjvallarino:
The `selectSubCompany` did not actually receive a `subCompanyName` parameter, even though we were relying on it for some of our tests. It's now being handled as an optional argument, keeping the previous behavior when it's not provided.

@Rise-Vision/apps @santiagonoguez @stulees please review


test/e2e/pages/commonHeaderPage.js
```diff
@@ -87,12 +87,18 @@
...
-    this.selectSubCompany = function() {
+    this.selectSubCompany = function(subCompanyName) {
       helper.clickWhenClickable(profilePic, 'Profile Picture');
       helper.clickWhenClickable(selectSubcompanyButton, 'Select Sub Company Button');
       helper.wait(selectSubcompanyModal, "Select Subcompany Modal");
       helper.waitDisappear(selectSubcompanyModalLoader, "Load Companies");
-      selectSubcompanyModalCompanies.get(0).click();
+
+      if (subCompanyName) {
+        element(by.xpath('//div[p/strong[contains(text(), "' + subCompanyName + '")]]')).click();
```
alex-deaconu: @fjvallarino do you think it would be useful to search for the Name via the Search Box? In case there's multiple pages. Thanks.

fjvallarino: @alex-deaconu yeah, it would make it more reliable on concurrent builds if a few people are working at the same time (especially if we create sub companies with unique names)

alex-deaconu: @fjvallarino sounds good, thanks for addressing!

PR #950, opened 2019-04-11
Refresh page to ensure User gets deleted

alex-deaconu:
Check for username rather than email


alex-deaconu:
Sometimes jenkins1 gets stuck and not deleted from jenkins company. This is because a warning dialog is open when the deletion tries to occur. Refreshing the page gets rid of the dialog.

Also, ensure that the Username is checked and not the email. We have some tests that fiddle with the email field value.

@fjvallarino @Rise-Vision/apps please review. Thanks!

PR #951, opened 2019-04-16
Allow configuration of useAnchor parameter for linker

alex-deaconu:
We need to use `true` for the RVA. Setting `false` for Apps to enable help center link anchors.

Staged at https://apps-stage-1.risevision.com


dist/js/components/segment-analytics.js
```diff
@@ -5,9 +5,11 @@
   angular.module("risevision.common.components.analytics", [])
 
   .value("SEGMENT_API_KEY", "AFtY3tN10BQj6RbnfpDDp9Hx8N1modKN")
+    .value("GA_LINKER_USER_ANCHOR", true)
```
ezequielc: @alex-deaconu I think it should be spelled GA_LINKER_USE_ANCHOR (USE instead of USER), right?

alex-deaconu: @ezequielc you're right, will fix before merging. Thanks!

alex-deaconu:
@ezequielc that is correct re: PR for RVA. Ideally we would default to `false` and set `true` in RVA, but I think this is ok too. Thoughts? Thanks.

ezequielc:
@alex-deaconu This is good with me. 👍 Thanks!

PR #952, opened 2019-04-16
Deprecate unused components to reduce requests

alex-deaconu:
Remove shopping cart, system messages
Remove subscription status check for priority support


olegrise:
@alex-deaconu not sure if we should remove the shopping cart. Store is not fully deprecated yet. We might not be able re-build it using the latest CH versions. What do you think?

alex-deaconu:
@olegrise my plan is to update Store to reflect these changes. Thoughts? Thanks.

olegrise:
@alex-deaconu that works. Thanks.

alex-deaconu:
Screen Shot 2019-04-16 at 4 16 44 PM

Staged at:
https://apps-stage-1.risevision.com
https://store-stage-0.risevision.com

PR #953, opened 2019-04-17
All Modals will not close on Esc or click outside

alex-deaconu:
Configure via default settings for
Clean out modal configuration overrides

Remove deprecated scss files
Remove intercom references and initialization


alex-deaconu:

PR #961, opened 2019-04-29
Make Apps e2e test cases independent in order to be able to run multiple builds in parallel

fjvallarino:
The main changes here are:

- When creating a sub company, append the stage environment to its name
- Apply the same logic on sub company selection (it also retries if the sub company list comes up empty, since sometimes because of caching or other causes Core's response does not return the expected result)
- When creating a sub company, if another with the same name exists, delete it first. This is done to avoid issues with previous test runs that may have had Subscribed/Started a Trial. It also helps avoiding multiple dangling copies of previously failed tests
- Adds several minor improvements to make tests more reliable

@santiagonoguez @stulees @Rise-Vision/apps please review


olegrise:
@fjvallarino looks good to me 👍

alex-deaconu:
@fjvallarino looks good 👍 . Thanks!

PR #963, opened 2019-05-01
Add translations for licensing expired modal

stulees:


olegrise:
@stulees why do we keep adding translations for things? I thought we decided that we are not going to do that any more because we only focus on IT Clay in North America.

stulees:
@olegrise That's true. We originally spoke to this in reference to component instance labels as it was going to be detrimental for creative creating templates if common-header was involved. However, aside from component labels, we've continued to add translations for consistency in Apps development. I'm not sure if/when we should stop this entirely, I'll raise this in standup tomorrow, thanks.

olegrise:
@stulees sounds good. Thanks!

PR #964, opened 2019-05-01
Add translation for rise-text component

olegrise:
@andrecardoso please review.


olegrise:
@andrecardoso this translation is used in Apps [here](https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/partials/template-editor/attribute-list.html#L5). Here is the screenshot:
fix translation

PR #967, opened 2019-05-07
Image component base

fjvallarino:
@santiagonoguez @stulees please review

@Rise-Vision/apps FYI


src/scss/sections/components/image.scss
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+.image-component-list {
+  height: calc(var(--image-component-body-height) - var(--image-component-list-button-bar-height));
+  overflow-x: hidden;
+  overflow-y: auto;
+
+  .image-component-list-empty {
+    color: rgb(153, 153, 153);
```
santiagonoguez: @fjvallarino do we have global color constants for this and other values ?

fjvallarino: @santiagonoguez not for that specific one (I just added it), but we do for the others. Updated

PR #972, opened 2019-05-15
Collect industry for sub-company

olegrise:
Fix for https://github.com/Rise-Vision/rise-vision-apps/issues/951

The `createSubCompany` [helper method](https://github.com/Rise-Vision/common-header/compare/feature/subcompany-collect-industry?expand=1#diff-8f6b438fd95e2f16e7edef3ef461c650R115) is only used in Apps tests for testing templates. It's not used anywhere in CH.

@andrecardoso please review.
@alex-deaconu I wonder if you could review as well. @ezequielc FYI.


olegrise:
https://trello.com/c/JRHfcj5S/43-3-issue-951-user-not-asked-to-confirm-industry-when-navigating-to-sub-company

alex-deaconu:
@olegrise looks good 👍 . Thanks!

andrecardoso:
@olegrise the changes look good, but I'm missing the part that implements this part: "Note: a user that belongs to the Rise Vision Company should not be asked to confirm industry of a sub-company".

I think we need an extra if in the handler to skip collecting the industry if the parent company is Rise Vision Company. Thoughts?

olegrise:
> @olegrise the changes look good, but I'm missing the part that implements this part: "Note: a user that belongs to the Rise Vision Company should not be asked to confirm industry of a sub-company".
> 
> I think we need an extra if in the handler to skip collecting the industry if the parent company is Rise Vision Company. Thoughts?

@andrecardoso we already have this `if` implemented
```
if (!userState.isUserAdmin()) {
    return;
}
```
[here](https://github.com/Rise-Vision/common-header/pull/972/files#diff-6b488e12aa3b7800808879d380410985R109).

olegrise:
@andrecardoso Good catch re. "Rise user" use case. I fixed and restaged.

PR #974, opened 2019-05-16
Account for space of Add Selected button

fjvallarino:
@santiagonoguez @stulees please review


fjvallarino:
@santiagonoguez yes, in this PR: https://github.com/Rise-Vision/rise-vision-apps/pull/977

The mockups for Storage Selector do now show an Upload Images button when you're inside a folder, but it should actually be visible. This PR adjusts the size of the bottom section to account for this new element

PR #981, opened 2019-05-27
Round trial days up on subscription status

alex-deaconu:
Minor version increase


alex-deaconu:
Fix is staged on stage-2. Minor version bump just to reset count.

PR #982, opened 2019-05-29
Add rise-image label

stulees:


stulees:
@santiagonoguez I was going to leave it since we have existing old version example financial template presentations that use that label, but I agree it's not required to keep it. I removed it.

PR #984, opened 2019-05-30
Styles for image component duration setting

stulees:
- Implemented on PR https://github.com/Rise-Vision/rise-vision-apps/pull/1002
- Validate with https://apps-stage-8.risevision.com/templates/edit/c746a8bf-69da-42db-bf65-9eb7a75f587f?cid=dd474bee-b237-46e3-aa20-98e975679773


stulees:
@santiagonoguez I see that now. Hm, tricky, as the height calculation is based on the 3 sections fixed heights:  duration, image list, and action buttons. The duration part is set to not show if less than 1 item. This is likely going to require JS to dynamically update height value.

fjvallarino:
@stulees maybe you can conditionally apply classes with `ng-class` and update the height in a similar way to how we do here: https://github.com/Rise-Vision/common-header/blob/master/src/scss/sections/template-editor-layout.scss#L35

stulees:
@fjvallarino 👍 good call, will do

stulees:
@fjvallarino @santiagonoguez Changes made, you can validate again with https://apps-stage-8.risevision.com/templates/edit/c746a8bf-69da-42db-bf65-9eb7a75f587f?cid=dd474bee-b237-46e3-aa20-98e975679773 as the Apps PR also applies the change - https://github.com/Rise-Vision/rise-vision-apps/pull/1002

src/scss/sections/components/image.scss
```diff
@@ -138,6 +144,37 @@
...
 }
 
+.image-component-list-duration {
+  height: var(--image-component-list-duration-height);
+  padding: 10px 20px 0;
+  background: $white;
+
+  & .form-control {
+    height: 40px;
+    color: #020621;
```
santiagonoguez: @stulees I think it would be better to use the $gray-dark constant here

src/scss/sections/components/image.scss
```diff
@@ -138,6 +144,37 @@
...
+    border: 1px solid #999999;
```
santiagonoguez: @stulees same here, we have $mid-gray

src/scss/sections/components/image.scss
```diff
@@ -138,6 +144,37 @@
...
+    border-right: 0;
+    box-shadow: none;
+    transition: none;
+
+    &:focus {
+      border-color: #107dda;
```
santiagonoguez: @stulees I didn't find a constant here, but maybe we should create one

src/scss/sections/components/image.scss
```diff
@@ -138,6 +144,37 @@
...
+
+    &:focus + .input-group-addon {
+      border-color: #107dda;
+    }
+  }
+
+  & .input-group-addon {
+    font-size: 12px;
+    color: #020621;
+    background-color: #f2f2f2;
```
santiagonoguez: @stulees $lighter-gray

stulees:
@santiagonoguez Good call, I was following the new style guide for colours and applying them directly, will leverage the bootstrap and our own variables next time. Thanks!

PR #990, opened 2019-06-06
Fix/folder row breaks

santiagonoguez:
These CSS properties fix the folder break issue. It can be validated by going into the long folder in storage here:

https://apps-stage-9.risevision.com/templates/edit/0275781b-b4a4-4c58-aa0a-75992ae2d995?cid=cf85e5c4-9439-40ce-94d6-e5ea6ea2411c

Before applying these styles it would look like:

staging

And once fixed it looks like:

fixed


stulees:
@santiagonoguez Nice one 👍

PR #995, opened 2019-06-12
Template Editor modals refactoring / New style

ezequielc:
I refactored styling on Template Editor modals so that we can have a single modal class and reuse it.

Also, I introduced `madero-style` class as a wrapper for the new style guide changes, so that we can make elements reusable.
For example, wen we apply `madero-style` to a `bnt-default` button, it will use the new style.
My idea is to gradually add `madero-style` to other areas of the app, so that we can gradually transition to new style, while avoiding specific classnames as we are currently doing (template-editor-message-box, etc)

@alex-deaconu Please review. Thanks


alex-deaconu:
@ezequielc looks good, I would do a minor version bump for this release. Thanks!

ezequielc:
@andrecardoso @olegrise @stulees @fjvallarino @santiagonoguez FYI, please take a look at this approach for implementing the new style-guide changes. I think it is a good way to create reusable and isolated CSS changes. Let me know if you have any feedback. Thanks!

PR #1001, opened 2019-06-19
Fix issue 1015 incorrect plan status

olegrise:
@andrecardoso please review.
@alex-deaconu I wonder if you could take a look as well.

BTW, this issue occurs only when subscription was purchased by parent for a sub-company. It does not occur when subscription is shared by parent company.


olegrise:
Staged on stage-3

alex-deaconu:
@olegrise looks good! Thanks for fixing 👍

olegrise:
Fix for https://github.com/Rise-Vision/store-server/issues/1015

PR #1002, opened 2019-06-19
Feature/volume pricing

tejohnso:
Changes for volume pricing (go-live is Monday).
The individual commits should be reasonably easy to follow.
I left the build changes and test changes in separate commits. 
I'll update the bower version before merge.

Live on [stage 6](https://apps-stage-6.risevision.com)


src/js/components/plans/ctr-plans-modal.js
```diff
@@ -1,174 +1,248 @@
...
-    function _getPlansDetails() {
-      $loading.start("plans-modal");
-
-      return plansFactory.getPlansDetails()
-        .then(function (resp) {
-          $scope.plans = resp;
-        })
-        .finally(function () {
-          $loading.stop("plans-modal");
+  .directive("rvOn", function () {
```
alex-deaconu: @tejohnso can this be moved to a different file? Thanks.

dist/css/rise.css
```diff
@@ -5689,23 +5689,24 @@ button.close {
 /* FontAwesome
    ========================================================================== */
 /*!
- *  Font Awesome 4.7.0 by @davegandy - http://fontawesome.io - @fontawesome
+ *  Font Awesome 4.1.0 by @davegandy - http://fontawesome.io - @fontawesome
```
alex-deaconu: @tejohnso your dependencies seem to be out of date here. It's causing `dist` differences despite not having any changes in bower or scss. Same goes for the font files below. Wondering if you could review and fix. Thanks.

src/js/components/plans/svc-plans-factory.js
```diff
@@ -104,9 +131,9 @@
       productCode: "d521f5bfbc1eef109481eebb79831e11c7804ad8",
       proLicenseCount: 0
     }])
-    .factory("plansFactory", ["$q", "$log", "$modal", "$templateCache",
+    .factory("plansFactory", ["$q", "$log", "$modal", "$templateCache", "getCompany",
```
alex-deaconu: @tejohnso I don't think this import is being used. Thanks.

src/js/components/plans/ctr-plans-modal.js
```diff
@@ -1,174 +1,248 @@
...
+          .then(function (industry) {
+            $scope.pricingComponentDiscount = volumePlan.discountIndustries.includes(industry);
+          });
+      }
 
-    $scope.isTrialExpired = function (plan) {
-      return plan.statusCode === "trial-expired";
-    };
+      function _getIndustry() {
+        var cid = window.top.location.href.match(/cid=([a-z0-9-]+)/);
```
alex-deaconu: @tejohnso you should be using `userState.getCopyOfSelectedCompany()` for this. Thanks.

src/js/components/purchase-flow/controllers/ctr-purchase-modal.js
```diff
@@ -1,29 +1,25 @@
 angular.module("risevision.common.components.purchase-flow")
 
 .value("PURCHASE_STEPS", [{
-  name: "Subscription Details",
-  index: 0,
-  formName: "reviewSubscriptionForm"
-}, {
```
alex-deaconu: @tejohnso curious, instead of removing this first page, and closing and opening the Plans Modal, why not just make the Plans Modal the first page here? Thanks.

src/js/components/plans/ctr-plans-modal.js
```diff
@@ -1,174 +1,248 @@
...
-    };
+        return getCompany(companyId).then(function (company) {
+          return company.companyIndustry;
+        });
+      }
 
-    $scope.isFree = function (plan) {
-      return plan.type === "free";
-    };
+      function _getPlansDetails() {
```
alex-deaconu: @tejohnso I don't know the logic of your new modal, but I'm pretty sure a lot of this code can be removed. Thanks.

src/js/config/prod.js
```diff
@@ -35,6 +35,7 @@
     .value("APPS_URL", "https://apps.risevision.com")
     .value("CHARGEBEE_TEST_SITE", "risevision-test")
     .value("CHARGEBEE_PROD_SITE", "risevision")
+    .value("CHARGEBEE_PLANS_USE_PROD", "true")
```
alex-deaconu: @tejohnso to confirm, when common-header is built, we use this (`prod.js`) config. What this means is that in Apps you're always using this value. If you wanted to have this value different based on staging/prod, you should update apps config files as well. Thanks.

src/templates/components/plans/plans-modal.html
```diff
@@ -1,197 +1,17 @@
...
-            common-header.plans.individual-licenses
-          </a>
-
-          <!-- Trial remaining days -->
-          <p id="trial-days-remaining" class="small u_margin-sm-bottom text-subtle" ng-show="isCurrentPlan(plan) && isOnTrial(plan)" translate
-            translate-values="{ count: currentPlan.trialPeriod }">common-header.plans.days-left-trial</p>
-          <!-- Trial expired notice -->
-          <p class="small u_margin-sm-bottom text-danger" ng-show="isCurrentPlan(plan) && isTrialExpired(plan)" translate>common-header.plans.trial-expired</p>
+<div>
+  <button type="button" style="display: block; width: 100%; text-align: center; padding: 0.5em 0.5em 0 0; background: none; border: none; text-align: right" ng-click="dismiss()" aria-hidden="true">
```
alex-deaconu: @tejohnso wondering if you could clean out the inline styles from this partial. Thanks!

src/templates/plan-banner.html
```diff
@@ -33,13 +33,6 @@
     </div>
   </div>
 
-  <div class="alert alert-plan plan-active text-right" ng-show="getVisibleBanner() === 'subscribed'">
-    <div class="u_margin-right">
-      <span class="font-weight-bold">{{plan.name}} Plan</span>
-      <a href="#" ng-show="!isEnterpriseSubCompany()" ng-click="showPlans()" class="ml-3 btn btn-xs btn-primary">Change Plan</a>
```
alex-deaconu: @tejohnso if the plan is to no longer allow users to "Change Plan", there is another usecase above (Line 9) that may need to be reviewed. Thanks.

alex-deaconu:
@tejohnso reviewed the code, my comments are inline. I have yet to review the apps PR or the staged version.

You will also need to bump the version in the `bower.json`/`package.json` files. I would suggest bumping a minor vesion to `3.9.0`.

Also, what is the plan for staging/releasing this in the RVA?

Thanks.

tejohnso:
@alex-deaconu Most issues addressed. See latest commits.

Not addressed: 

 - Unused code
 - Reorganizing pricing component modal into purchase flow modal
 - RVA

I will work on the unused code now. On my first attempt the plan banner stopped working properly. It began showing the Show Plans option even when subscribed.

Bringing pricing component modal into the purchase flow modal requires some UI design consideration and is out of scope for now.

RVA is a lower priority and I'll tackle that later. Plan is to redirect to pricing page rather than bringing the component into RVA. But I haven't changed RVA in years and I hear it doesn't have automated deployment. It might fall onto you to make that change :)

src/templates/plan-banner.html
```diff
@@ -1,15 +1,5 @@
 <!-- Common Header Plan Info -->
 <div class="hidden-xs">
-  <div id="parent-plan-banner" class="alert p-2 mb-0 bg-success text-right" ng-show="getVisibleBanner() === 'parent'">
```
alex-deaconu: @tejohnso actually, I'm wondering what the purpose of these changes is in the Plan Banner. If we are not showing a banner when they are Subscribed, then you probably have to remove the below (`parent-purchased-plan-banner`) as well. Thanks.

alex-deaconu:
@tejohnso thanks for the fixes, looks good!

I'm good re: reorganizing modals, I realize it may require more refactoring than it's worth.

Okay re: RVA.

Thanks!

src/templates/components/plans/plans-modal.html
```diff
@@ -1,197 +1,17 @@
...
-          </div>
-          <h4 class="text-center" style="column-span: all;"><a href="https://www.risevision.com/pricing" target="_blank">Learn More About Key Features</a></h4>
-        </div>
-      </div>
-      <div id="planFooter"></div>
-    </div>
+  <div id="pricingComponentContainer">
+    <pricing-component
+      ng-attr-apply-discount="{{pricingComponentDiscount ? '' : undefined}}"
+      ng-attr-prod-env="{{useProductionChargebeeData ? '' : undefined}}"
```
olegrise: @tejohnso isn't `prod-env` suppose to be a boolean value?

src/templates/components/plans/plans-modal.html
```diff
@@ -1,197 +1,17 @@
...
-          <div class="planFeature">
-            <p class="featureTitle">Display On/Off Control</p>
-          </div>
-          <h4 class="text-center" style="column-span: all;"><a href="https://www.risevision.com/pricing" target="_blank">Learn More About Key Features</a></h4>
-        </div>
-      </div>
-      <div id="planFooter"></div>
-    </div>
+  <div id="pricingComponentContainer">
+    <pricing-component
```
olegrise: @tejohnso do I understand it right that `pricing-component` is loaded by Apps but rendered here in CH?

src/templates/components/plans/plans-modal.html
tejohnso: Yes, ng-attr will omit the attribute altogether if the expression result is `undefined` and include it otherwise.

src/templates/components/plans/plans-modal.html
tejohnso: Yes, Apps controls the loading because it owns `index.html`
But the component replaces the plans-modal which comes from CH.

Willing to move things around if desired. I think the pricing-component.mjs can go into CH as well, but I understand the plan is to merge both anyway.

src/templates/components/plans/plans-modal.html
olegrise: Aah, makes sense. It wasn't clear from the code that the intention is to omit the attribute. Thanks.

src/templates/components/plans/plans-modal.html
olegrise: Hmm. I agree that if we merge Apps and CH, then it won't matter, but I don't know if it will ever happen. @alex-deaconu curious what you think.

src/js/components/plans/svc-plans-factory.js
```diff
@@ -116,19 +143,14 @@
           _plansByType.free, _plansByType.starter, _plansByType.basic, _plansByType.advanced, _plansByType.enterprise
         ];
 
-        _factory.showPlansModal = function (warningText) {
```
alex-deaconu: @tejohnso to confirm, this parameter is used to display warnings:
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/scripts/editor/services/svc-editor-factory.js#L429
https://github.com/Rise-Vision/rise-vision-apps/blob/master/web/partials/displays/display-details.html#L48

Are we good with this no longer being available? If so, can you remove the warnings from Apps?

Thanks.

tejohnso: As discussed, we'll have to bring this up for UI review.

PR #1004, opened 2019-07-04
Feature/new component labels

santiagonoguez:
I added both video and RSS labels to avoid an extra simple PR in the future.


fjvallarino:
@santiagonoguez thanks!

PR #1012, opened 2019-07-18
Update phone number for user rather than company object

alex-deaconu:


alex-deaconu:
Staged at https://apps-stage-2.risevision.com

@andrecardoso thanks for the review!

PR #1013, opened 2019-07-18
Add directive that checks and sets CH height

alex-deaconu:
Used for Editor and Template Editor
It sets the height based on the actual value
Refresh values on Company update/change and Plan change


alex-deaconu:
Fix for https://github.com/Rise-Vision/rise-vision-apps/issues/1080. Instead of using states to determine the height of the common-header arbitrarily, this directive checks the actual rendered height and updates the CSS variable.

Staged at https://apps-stage-2.risevision.com.

@fjvallarino @andrecardoso please review. Thanks!

PR #1014, opened 2019-07-18
Fix for issue 1054 - industry modal opens after user registration

olegrise:
Fix for https://github.com/Rise-Vision/rise-vision-apps/issues/1054.
Staged on stage-3. Here is the corresponding Apps PR https://github.com/Rise-Vision/rise-vision-apps/pull/1119

There were 2 problems:
1. The main problem was that at the end of User Registration the `updateCompany` method was called. As a result the apps state and selected company were reloaded immediately without waiting for `updateCompany` to finish. As a result the selected company was loaded from DB before industry was saved. The fix this situation is [here](https://github.com/Rise-Vision/common-header/compare/fix/issue-1054-industry-modal-opens-after-user-registration?expand=1#diff-2ec175c12c8011573203566f05849457R1612).
2. The secondary problem is that `risevision.company.selectedCompanyChanged` event is sent twice after User Registration and before industry is saved. This caused "flickering" of the Industry Collection Modal. It appeared for a moment when `risevision.company.selectedCompanyChanged` was received and then disappeared then apps state was reloaded. To fix this, I added a check for presence of the `"common.auth"` in the state name. There is more than one state that start with `"common.auth"` and from what I can tell we don't need to show ICP modal for any of these states. 

@andrecardoso please review.
@alex-deaconu FYI


src/js/services/svc-company-icp.js
```diff
@@ -103,6 +103,12 @@ angular.module("risevision.common.header")
       };
 
       var _checkIcpCollection = function () {
+
+        //make sure user registration process is complete
+        if ($state.current.name.indexOf("common.auth") !== -1) {
```
alex-deaconu: @olegrise curious, what `common.auth...` states does this apply to? I'm assuming there's more than 1 since you're doing this comparison. Thanks.

alex-deaconu:
@olegrise just one question inline; this is more for my information. Otherwise PR looks good. Ideally however I would refactor and create a registration API where the User and Company can be created/updated all at once instead of this complicated flow.

Thanks!

src/js/services/svc-company-icp.js
olegrise: @alex-deaconu I only saw `common.auth.unregistered`, but I though it's safer to filter out all `common.auth.*` states because it doesn't make sense to show Industry Collection Modal when CH is in any of those states. I sort of copied this logic from [here](https://github.com/Rise-Vision/common-header/blob/e8e8e7954df9e1da228712a9ad5fa8a6973bd07d/src/js/components/userstate/app.js#L193). Do you think it's better to specifically check for `common.auth.unregistered`?

olegrise:
@alex-deaconu I agree that refactoring registration and calling Core API only once would be the best option. It was done the way it is because of security concerns, but maybe it's time to review this flow.

src/js/services/svc-company-icp.js
alex-deaconu: @olegrise no, I'm fine with leaving as is, I was just checking the code and was wondering if there were other states that appeared in your testing. Thanks.

olegrise:
@alex-deaconu I spent some time thinking about the idea of refactoring registration. My main concern is that we keep changing the business logic around user/company registration - we add fiends, we remove fields, we start trial,  we don't start trial, etc. The current API and CH implementation even though not optimized but flexible enough to allow us to make these changes fairly easy. I decided to leave everything as is for now. Thanks.

alex-deaconu:
@olegrise I'm totally ok with that, I just wanted to mention that as a potential improvement. Definitely outside of the scope for this fix. Thanks.

PR #1015, opened 2019-07-19
Attach handler to scope for cleanup

alex-deaconu:


alex-deaconu:
Quick follow up PR for the release yesterday. It seems `$rootScope.$on` handlers are not removed when directives are unloaded. Instead of removing the handlers manually, I've opted to use `$scope.$on`.

@andrecardoso @fjvallarino please review. Thanks!

alex-deaconu:
@fjvallarino thanks!

Note: I will release and create a tag, and this can be attached to the next CH release (ie. I won't release apps specifically for this change).

